Issue #1, opened 2023-05-26
ConnectGuardians.tsx displays wrong value


EthnTuttle:
```tsx
        <CopyInput
          value={process.env.REACT_APP_FM_CONFIG_API || ''}
          size='lg'
          buttonLeftIcon={<Icon as={CopyIcon} />}
        />
```

We currently use the UI enviroment variable as the connection. However, this is incorrect and should be the `FM_API_URL` passed into the Coordinators config for fedimintd.

EthnTuttle:
https://github.com/fedimint/ui/issues/19
Related.

PR #23, opened 2023-06-03
Use API for Peer Id

EthnTuttle:
The api returns `our_current_id` for `get_verify_config_hash`, so we use that. Since PeerId is part of ConsensusConfigGenParams too, we can determine which peer we are/are not.


EthnTuttle:
Addresses one of the issues in #1

okjodom (approved):

PR #25, opened 2023-06-07
Build Gateway UI with gateway API

okjodom:
- Cleans up a bunch of mocks in gateway UI
- Simplifies offering of gateway UI mvp (_and we'll look for design guidance on the experience to build from here on_)
- Simplifies API definition and restructures app to follow `guardian-ui` source organization
- Builds and uses gateway api `/info`
- Builds and uses gateway api `/address`

Replaces https://github.com/fedimint/fedimint/pull/2592


docker-compose.yml
```diff
@@ -4,9 +4,10 @@ services:
   fedimintd_1:
     image: fedimint/fedimintd:master
     environment:
+      - FM_PASSWORD=pass
```
okjodom, line 7: Fedimintd containers should not require password. Rather, the password should be provided by the guardians via Setup UI.
```
// without FM_PASSWORD
fedimintd exits with code 1. says FM_PASSWORD env var required
```

Need to fix this, for the setup UI to work

```
// setup failure at DKG
Unable to write to data dir Os { code: 17, kind: AlreadyExists, message: "File exists" }
```

apps/gateway-ui/src/GatewayApi.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// GatewayApi is an implementation of the ApiInterface
+export class GatewayApi implements ApiInterface {
+  private baseUrl: string | undefined = process.env.REACT_APP_FM_GATEWAY_API;
+  private password = process.env.REACT_APP_FM_GATEWAY_PASSWORD || '';
+
+  fetchInfo = async (): Promise<GatewayInfo> => {
+    try {
+      const res: Response = await fetch(`${this.baseUrl}/info`, {
```
okjodom, line 42: need to fix CORS on gateway api call

okjodom:
**Some Notes**:
- Guardian setup ui is currently broken at DKG
- Developing gateway UI requires a running federation to connect to. This experience currently directs the dev towards quickly setting up a federation via the fed setup-ui to get the connection string
  - Great if fed setup experience is simple and quick
  - Great if fed setup UI is working (I like that this forces gateway ui dev to set up fed)
  - Poor dev experience if setup Ui is broken! will block dev of gateway UI

docker-compose.yml
okjodom: fixed by https://github.com/fedimint/fedimint/pull/2587/files. I just had to clear my docker images and redownload to get the latest one which includes fix

garyray-k: We could probably update the command to `Docker-compose pull && Docker-compose up`. Then it would always grab the latest. Unless we peg to a commit hash.

justinmoon (approved):

PR #29, opened 2023-06-10
1 connectguardianstsx displays wrong value

EthnTuttle:
Saves `our_current_id` from the API to GuardianContext. 
Then uses that to filter and display `api_url` from `peers`.

Previously, it was displaying the url from the ENV VAR `REACT_APP_FM_CONFIG_API` which may not always be the same.


EthnTuttle:
Closes #1

okjodom (approved):

PR #30, opened 2023-06-10
build: adds clear-federation-data to mprocs

EthnTuttle:
Useful for when you need to clear the federation data and restart a setup.

### DIFF
--- a/mprocs.yml
+++ b/mprocs.yml
@@ -1,20 +1,23 @@
 procs:
   run-ui-federation:
-    shell: docker compose up 
+    shell: docker compose up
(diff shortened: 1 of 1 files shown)


okjodom (approved):

okjodom:
Neat. Tried to use this for dev and realized
1. I need `clear-federation-data` and `teardown-ui-federation` to be in the same process so I don't clear data without teardown
2. I need sudo to remove `fm_*` dirs

Refactored this definition and ended with this mprocs def: https://github.com/fedimint/ui/pull/31

Also runs a gateway-ui instance
image

PR #36, opened 2023-06-16
Feature Stepper Component

IgboPharaoh:
This pr introduces a stepper component that shows the specific point or step a user is at during the setup process of a federation.
 
The stepper was also made to work with the respective routes that guardians can take when a federation is being set up either as a Leader or as a Follower.

Figma file: https://www.figma.com/file/ioeAVbWBc3ow2RaZETatM7/Fedimint?type=design&node-id=497-19323&t=ypN1cLAysc1PRU2y-0

Please review @okjodom  @EthnTuttle 

Pictorial representation of the stepper:
Screenshot 2023-06-16 at 06 22 25

Screenshot 2023-06-16 at 06 23 01


apps/guardian-ui/src/components/CustomStepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { StepState } from '../types';
+
+interface StepProps {
+  text: string;
+  state: StepState;
+  margin?: string;
+}
+
+const Step: React.FC<StepProps> = ({ text, state, margin = '-48px' }) => {
+  const active = state === StepState.Active;
```
okjodom, line 13: nit: I'd simplify and just use `StepState` directly instead of creating consts here

apps/guardian-ui/src/components/CustomStepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const inActive = state === StepState.InActive;
+  const completed = state === StepState.Completed;
+
+  const colorState = (
+    inActiveColor: string,
+    activeColor: string,
+    completedColor: string
+  ) => (inActive ? inActiveColor : active ? activeColor : completedColor);
+
+  const rectangleState = () =>
```
okjodom, line 23: nit: this is only used once, so maybe we should just inline the logic instead of creating this function named `rectangleState`?

apps/guardian-ui/src/components/CustomStepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+        whiteSpace='nowrap'
+        mr={margin}
+      >
+        {text}
+      </Text>
+    </Flex>
+  );
+};
+
+const InitialStep: React.FC<Pick<StepProps, 'state'>> = ({ state }) => {
```
okjodom, line 77: why is the initial step special? Can we avoid having a special initial step?

apps/guardian-ui/src/components/CustomStepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+    </Flex>
+  );
+};
+
+interface StepperProps {
+  step: number;
+  isHost: boolean;
+}
+
+const CustomStepper: React.FC<StepperProps> = ({ step, isHost }) => {
```
okjodom, line 129: nit: `SetupProgress` is probably a more descriptive name

okjodom: export here, instead of default export below

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -19,6 +19,7 @@ import { ConnectGuardians } from '../components/ConnectGuardians';
 import { RunDKG } from '../components/RunDKG';
 import { VerifyGuardians } from '../components/VerifyGuardians';
 import { SetupComplete } from '../components/SetupComplete';
+import CustomStepper from '../components/CustomStepper';
```
okjodom, line 22: we probably want to avoid default exports

apps/guardian-ui/src/types.tsx
```diff
@@ -13,6 +13,12 @@ export enum PeerConnectionStatus {
   Disconnected = 'Disconnected',
 }
 
+export enum StepState {
```
okjodom, line 16: move to setup/types.tsx

okjodom:
I left a couple nits related to naming and code conventions. But the experience is neat and informative to the users

- progress component heavily uses colors and states. We should consider linking this to theme values
- do our designs spec how the progress experience looks like in mobile? If not, please leave a note in the figma file
- can you link to the figma file in description for reference

IgboPharaoh:
> I left a couple nits related to naming and code conventions. But the experience is neat and informative to the users
> 
> * progress component heavily uses colors and states. We should consider linking this to theme values
> * do our designs spec how the progress experience looks like in mobile? If not, please leave a note in the figma file
> * can you link to the figma file in description for reference

Will attend to all corrections and make fixes where necessary

apps/guardian-ui/src/components/CustomStepper.tsx
IgboPharaoh: Based on the Figma design and the design structure used to achieve the stepper, we need an initial step state. Making it to be optional from the design structure of the step component, will make the initial step to be disjointed from the rest of the design

okjodom (approved):

PR #38, opened 2023-06-20
refactor: add type safety to rpc methods

### LINKED ISSUE #33: type guard API methods
We can type guard this using `'status' | 'set_password' | etc. etc.` so only certain strings can be used when making the call. It then also serves as a one stop point of reference for what RPC methods are supported. I believe that to be especially important if there's ever a time when we export this API as a package/library for use in other apps.

_Originally posted by @EthnTuttle in https://github.com/fedimint/ui/pull/28#discussion_r1225980111_

okjodom:
closes #33

### DIFF
--- a/apps/guardian-ui/src/GuardianApi.ts
+++ b/apps/guardian-ui/src/GuardianApi.ts
@@ -18,9 +18,16 @@ export interface SocketAndAuthInterface {
 }
 
 interface RpcInterface {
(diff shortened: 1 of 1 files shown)


okjodom:
@EthnTuttle, could you take a look at this PR?

EthnTuttle:
LGTM

EthnTuttle (approved):

PR #49, opened 2023-07-05
build: add guardian-ui docker

douglaz:


justinmoon (approved):
LGTM. This just creates a new docker image which serves guardian UI. Let's see if it works!

PR #50, opened 2023-07-05
Add ability to run daemons using nix

justinmoon:
* `just nix-guardian` will spin up 4 guardian servers & UIs which
    haven't run DKG yet
    * `just nix-gateway` will spin up a 3/4 federation which has run
    DKG, and a gateway UI connected to the CLN node
    * `just docker` will run the existing `mprocs.yml` file

    Whenever we want to update fedimint, we just change the git hash
    in `flake.nix`.

Kinda annoying how many scripts and configs I needed to add. Can probably trim them down a little.


justfile
```diff
@@ -0,0 +1,11 @@
+default:
```
okjodom, line 1: Perhaps we could use yarn/npm scripts instead of adding just tooling.

```
"scripts": {
    ...
    "nix-guardian": "./scripts/mprocs-nix-guardian.sh",
    "nix-gateway": "./scripts/mprocs-nix-gateway.sh",
    "mprocs": "mprocs -c ./mprocs.yml"
 }
```

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -63,6 +63,7 @@ export const FederationAdmin: React.FC = () => {
   const consensusVersion =
     versions?.core.consensus !== undefined ? `${versions.core.consensus}` : '';
 
+  console.log(status);
```
wbobeirne, line 66: 🙅 
```suggestion
```

justinmoon: Fixed

justfile
okjodom: Even more, I propose we replace the current dev scripts. They are not useful
https://github.com/fedimint/ui/pull/86/commits/24e753afa4109b43d0ba05e3fdb22c1f0c7da784

okjodom:
I noted some issues with cleanup of devimint-spawned processes, and `yarn nix-gateway` would get stuck with error `waiting waiting for ui/api startup`.

Checking in this initial pass to set up nix + devimint for development. We'll iterate as we try use it more for everyday dev

okjodom (approved):

justinmoon:
Yea I think devimint needs a little more work on killing it's child processes more consistently

PR #53, opened 2023-07-06
chore: debugging hacks

justinmoon:
don't merge


wbobeirne:
Can this be closed now that #55 is in?

Issue #65, opened 2023-07-11
Guardian UI: Use find / replace for FM_CONFIG_API env var instead of rebuilding project in docker


wbobeirne:
https://github.com/fedimint/ui/blob/master/Dockerfile.guardian-ui#L5-L7

This site should be able to run on really lean & weak hardware since it's just a basic web app, but doing the build can blow out memory on some smaller machines. A quick find / replace would be much more efficient than doing a full build. Should also improve start-up times dramatically when not using the default env var.

wbobeirne:
This was fixed by #72

PR #68, opened 2023-07-12
feat: added gateway-ui docker

douglaz:


.github/workflows/ci.yml
```diff
@@ -18,9 +18,15 @@ jobs:
...
         with:
           file: Dockerfile.guardian-ui
           push: true
           tags: fedimintui/guardian-ui:latest
+      - name: Build and push gateway-ui
+        uses: docker/build-push-action@v4
+        with:
+          file: Dockerfile.gateway-ui
+          push: true
+          tags: fedimintui/gateway-ui:latest
```
wbobeirne, line 32: Does it make sense to have 2 different workflows for these images, and to have them configure `paths` to reduce unnecessary runs or identical docker images? Something like:
```yml
# .github/workflows/guardian-ui.yml
on:
  push:
    branches:
      - master
    paths:
      - .github/workflows/guardian-ui.yml
      - ui/apps/guardian-ui/**
      - ui/packages/**
```
and
```yml
# .github/workflows/gateway-ui.yml
on:
  push:
    branches:
      - master
    paths:
      - .github/workflows/gateway-ui.yml
      - ui/apps/gateway-ui/**
      - ui/packages/**
```

Dockerfile.gateway-ui
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /app 
+COPY . /app
+RUN yarn install
+RUN yarn build
+RUN npm install -g serve
+# The following env variables can be given at runtime:
+# - PORT
+# - REACT_APP_FM_GATEWAY_API
+# - REACT_APP_FM_GATEWAY_PASSWORD
+CMD yarn build && serve -s apps/gateway-ui/build/
```
wbobeirne, line 11: Similar to #65, we should ideally just do a simple find / replace rather than a full rebuild. This is fine for now though, and that issue can tackle the find / replace for both.

wbobeirne (approved):
Works great for me. Going to do a quick follow-up with some Dockerfile optimizations built off of this.

PR #69, opened 2023-07-12
Config and List Gateways API

okjodom:
Adds admin apis for fetching client config and listing registered gateways.
- To fetch config, we need connection string
```ts
  const [config, setConfig] = useState<ConfigResponse>();
  api.config(connectionCode).then(setConfig).catch(console.error);
```

- To list gateways, we need config, from which we can determine the lightning module id
```ts
  const [gateways, setGateways] = useState<object>(['foo']);
  if (config) {
      for (const [key, val] of Object.entries(config.client_config.modules)) {
        if (val.kind === 'ln') {
          api
            .moduleApiCall<Gateway[]>(key, LightningModuleRpc.listGateways)
            .then(setGateways)
            .catch(console.error);
          return;
        }
      }
    }
```

### DIFF
--- a/apps/guardian-ui/src/GuardianApi.ts
+++ b/apps/guardian-ui/src/GuardianApi.ts
@@ -1,6 +1,7 @@
 import { JsonRpcError, JsonRpcWebsocket } from 'jsonrpc-client-websocket';
 import { ConfigGenParams, ConsensusState, PeerHashMap } from './setup/types';
 import {
+  ConfigResponse,
   ConsensusStatus,
   ServerStatus,
   StatusResponse,
@@ -20,7 +21,7 @@ export interface SocketAndAuthInterface {
 interface RpcInterface {
   call: <T>(
     method: SetupRpc | AdminRpc | SharedRpc,
-    params?: object | null
+    params?: unknown
   ) => Promise<T>;
   // TODO: Consider moving this to `SocketAndAuthInterface` as part of the authentication methods.
   clearPassword: () => void;
@@ -130,7 +131,7 @@ class BaseGuardianApi
 
   call = async <T>(
     method: SetupRpc | AdminRpc | SharedRpc,
-    params: object | null = null
+    params: unknown = null
   ): Promise<T> => {
     try {
       const websocket = await this.connect();
@@ -191,12 +192,14 @@ enum AdminRpc {
   fetchEpochCount = 'fetch_epoch_count',
   consensusStatus = 'consensus_status',
   connectionCode = 'connection_code',
+  config = 'config',
 }
 
 export interface AdminApiInterface extends SharedApiInterface {
   version: () => Promise<Versions>;
   fetchEpochCount: () => Promise<number>;
   connectionCode: () => Promise<string>;
+  config: (connection: string) => Promise<ConfigResponse>;
 }
 
 export class GuardianApi
@@ -331,5 +334,9 @@ export class GuardianApi
 
   connectionCode = (): Promise<string> => {
     return this.base.call(AdminRpc.connectionCode);
+  };
+
+  config = (connection: string): Promise<ConfigResponse> => {
+    return this.base.call<ConfigResponse>(AdminRpc.config, connection);
   };
 }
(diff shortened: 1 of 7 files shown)


apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -191,12 +200,22 @@ enum AdminRpc {
   fetchEpochCount = 'fetch_epoch_count',
   consensusStatus = 'consensus_status',
   connectionCode = 'connection_code',
+  config = 'config',
+  module = 'module',
+}
+
+export enum LightningModuleRpc {
+  listGateways = 'list_gateways',
```
EthnTuttle, line 208: Should we migrate this to shared types as well? Would these RPCs calls be part of gateway UI?

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -332,4 +351,13 @@ export class GuardianApi
   connectionCode = (): Promise<string> => {
     return this.base.call(AdminRpc.connectionCode);
   };
+
+  config = (connection: string): Promise<ConfigResponse> => {
+    return this.base.call<ConfigResponse>(AdminRpc.config, connection);
+  };
+
+  moduleApiCall = <T>(moduleId: string, rpc: ModuleRpc): Promise<T> => {
+    const method = `${AdminRpc.module}_${moduleId}_${rpc}`;
```
EthnTuttle, line 360: `moduleId` is a u16 in fedimint-core/src/api.rs, should this be `number`?

apps/guardian-ui/src/types.tsx
```diff
@@ -47,3 +47,53 @@ export interface Versions {
...
 }
+
+export enum ModuleKind {
+  Ln = 'ln',
+  Mint = 'mint',
+  Wallet = 'wallet',
+}
+
+interface FedimintModule {
+  config: string;
```
EthnTuttle, line 58: This seems like it will be an interesting one to handle on the front end. How does the string representation look? Is it human readable?

EthnTuttle (commented):
Just a few questions.

apps/guardian-ui/src/types.tsx
```diff
@@ -47,3 +47,53 @@ export interface Versions {
...
+  federation_id: string;
+  api_endpoint: Record<number, ApiEndpoint>;
+  modules: Record<number, FedimintModule>;
+  meta: MetaConfig;
+}
+export interface ConfigResponse {
+  client_config: ClientConfig;
+}
+
+// These types are shared with the gateway-ui
```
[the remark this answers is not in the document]
okjodom: I think that's a great plan, but, maybe we don't do it just yet. One reason is, we haven't settled on where we want the gateway to live. There's considerations for building gateway UIs like Thunderhub, and not as this standalone app

apps/guardian-ui/src/types.tsx
```diff
@@ -47,3 +47,53 @@ export interface Versions {
...
+  proportional_millionths: number;
+}
+
+export interface Gateway {
+  gateway_pub_key: string;
+  api: string;
+  fees: Fees;
+  mint_channel_id: number;
+  node_pub_key: string;
+  route_hints: object[]; // FIXME
```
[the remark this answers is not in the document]
okjodom: https://github.com/fedimint/ui/issues/80

apps/guardian-ui/src/GuardianApi.ts
okjodom: These are only in Guardian UI. Afaik

apps/guardian-ui/src/types.tsx
okjodom: Oh this is a hash string. We'd need to decode it to get front-end usable JSON.

Example of config for LN module: `b0fee25a8e3758c11244dfa892243283e620581f94a6a14cf516210685ea978a1872832d3592812c6353295bc2885b6a00000000000000000000000000000000dab5bffa`

apps/guardian-ui/src/GuardianApi.ts
okjodom: fixed

EthnTuttle (approved):

PR #73, opened 2023-07-12
fix: missing yarn build on dockerfile

douglaz:


wbobeirne (changes requested):
See #74 for the real fix

PR #76, opened 2023-07-13
Prune docker images

### LINKED ISSUE #57: Optimize Docker container size
https://hub.docker.com/layers/fedimintui/guardian-ui/ is current ~1k Mb
We should improve on this.

### LINKED ISSUE #70: Dockerfiles are copying and building all apps unnecessarily
Each `Dockerfile` for each app should only copy and build the app in question. Running `yarn install` in the root directory installs all of the dependencies necessary for all packages and apps, and running `yarn build` in the root directory builds all apps. We should only copy and build the apps and packages that we need per app.

wbobeirne:
Closes #70 & closes #57. This builds on top of #74, the only new commit is edd28ab32bd4991e3b6edff8111b7f86b855508a

### What This Does

Adapts the Dockerfile from https://turbo.build/repo/docs/handbook/deploying-with-docker to our Dockerfiles.

### Results

Drastic file size reduction in images

* `guardian-ui`
  * Before: **3.67gb**
  * After: **191.41mb**
* `gateway-ui`
  * Before: **3.68gb**
  * After: **193.25mb**


douglaz (approved):
LGTM, let's merge it and test with the other docker compose files

PR #78, opened 2023-07-13
ci: build arm64 images in addition to amd64

wbobeirne:
I was getting the following error when trying to use the `docker-compose.yml` in `fedimint/docker`:
```
❯ docker-compose up
[+] Running 0/1
 ⠹ guardian-ui Pulling
no matching manifest for linux/arm64/v8 in the manifest list entries
```
This should provide an arm64 image. We may need to add more in the future if people run into this issue, but this should cover most of us for now.

### DIFF
--- a/.github/workflows/gateway-ui.yml
+++ b/.github/workflows/gateway-ui.yml
@@ -29,3 +29,4 @@ jobs:
           file: Dockerfile.gateway-ui
           push: ${{ github.ref == 'refs/heads/master' }}
           tags: fedimintui/gateway-ui:latest
+          platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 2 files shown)


wbobeirne:
Yikes, this ballooned the builds from ~3 minutes to ~16 minutes. Couple of options here:
1. Force people to emulate `amd64` images by specifying `platform: linux/amd64` in docker-compose configs
2. Only run `arm64` builds on master, leave it to amd64-only in PRs
3. Parallelize the builds to cut down the time just a bit
4. Suck it up and have 16 minute CI 🤷

douglaz:
Option 2+3 seems okay

wbobeirne:
We've decided to just specify `platform: linux/amd64` in `docker-compose` where necessary instead.

PR #85, opened 2023-07-17
chore: more docker tags for better versioning

douglaz:


wbobeirne (commented):
Seems handy once we start tagging releases, but for now AFAICT this would just keep pushing `:master` tags since we only push on master for the moment. Have we figured out a UI release strategy yet?

wbobeirne (approved):
My mistake, missed the `sha` tag. All good.

PR #104, opened 2023-07-21
Improve guardian ui boot experience

### LINKED ISSUE #34: Guardian: Reset Setup State
We should look to clear or reset `SetupState`  persisted in local storage once the guardian UI transition from setup mode to Admin dashboard experience

okjodom:
Simplifies application state management in Setup and admin UIs with the results that:
- Setup UI has a perceived faster load since we establish socket connections earlier when booting the app.
  - On boot, we always check latest server state to determine experience flow into Setup vs Admin UI
- Guardians can smoothly navigate back and forth, and reload the pages at every single step of setup UI
- On transition to Admin experience, we clear all setup-state persisted in local storage

fixes #34

### DIFF
--- a/apps/guardian-ui/src/AppContext.tsx
+++ b/apps/guardian-ui/src/AppContext.tsx
@@ -7,50 +7,20 @@ import React, {
 } from 'react';
 import { GuardianApi } from './GuardianApi';
 import { formatApiErrorMessage } from './utils/api';
+import { ServerStatus } from './types';
 
-const LOCAL_STORAGE_KEY = 'guardian-ui-state';
-
-function makeInitialState(loadFromStorage = true): AppState {
-  let storageState: Partial<AppState> = {};
-  if (loadFromStorage) {
-    try {
-      const storageJson = localStorage.getItem(LOCAL_STORAGE_KEY);
-      if (storageJson) {
-        storageState = JSON.parse(storageJson);
-      }
-    } catch (err) {
-      console.warn('Encountered error while fetching storage state', err);
-    }
-  }
-
-  return {
-    experience: 'Setup',
-    ...storageState,
-  };
-}
-
-const initialState = makeInitialState();
-
-interface AppState {
-  experience: 'Setup' | 'Admin';
-}
-
-interface ApiState {
-  connected: boolean;
-  error: string;
-}
+type status = 'Loading' | 'Setup' | 'Admin' | 'Error';
 
 export interface AppContextValue {
   api: GuardianApi;
-  apiState: ApiState;
-  appState: AppState;
+  appState: status;
+  appError?: string;
   transitionToAdmin: () => void;
 }
 
 export const AppContext = createContext<AppContextValue>({
   api: new GuardianApi(),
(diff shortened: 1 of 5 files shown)


apps/guardian-ui/src/setup/SetupContext.tsx
```diff
@@ -229,6 +220,13 @@ export const SetupContextProvider: React.FC<SetupContextProviderProps> = ({
         configGenParams: state.configGenParams,
       })
     );
+
+    return () => {
+      // Clear local storage on setup complete.
+      // This happens when we transition to admin experience.
+      state.progress === SetupProgress.SetupComplete &&
+        localStorage.removeItem(LOCAL_STORAGE_SETUP_KEY);
+    };
```
wbobeirne, line 229: Do we want this in the useEffect teardown function? Seems to me like the useEffect should be an `if / else`:

```ts
if (state.progress === SetupProgress.SetupComplete) {
  localStorage.removeItem(...)
} else {
  localStorage.setItem(...)
}
```

apps/guardian-ui/src/AppContext.tsx
```diff
@@ -7,50 +7,20 @@ import React, {
...
-
-interface AppState {
-  experience: 'Setup' | 'Admin';
-}
-
-interface ApiState {
-  connected: boolean;
-  error: string;
-}
+type AppState = 'Loading' | 'Setup' | 'Admin' | 'Error';
```
wbobeirne, line 12: This is kind of a nomenclature nitpick, but `AppState` tends to be an object representing the full state of something. I would consider this more of a `status` since it's a string. I think that would provide some improved readability to call it a status, since it was previously an object, and setup continues to use an object for its `state` variable and `SetupState` type as well.

wbobeirne (commented):
Code review notes only, haven't tested yet in browser

apps/guardian-ui/src/setup/SetupContext.tsx
wbobeirne: To clarify, the way this is written now would cause the `removeItem` to run, but then immediately the useEffect would run with the new dependencies and re-write to `localStorage` again.

okjodom: > Do we want this in the useEffect teardown function? Seems to me like the useEffect should be an if / else

An `if` is sufficient. We only nead to clean up the local storage in the singular case when we got to `SetupComplete`, and are now tearing down `SetupContext`. 

> but then immediately the useEffect would run with the new dependencies and re-write to localStorage again.

Logically, we always (re)boot up to admin experience from this point onwards, thus we are _terminally tearing down `SetupContext`_ and this effect that watches state changes never runs again

Rewritten as an `if` instead of the `&&` shortcut

wbobeirne: Gotcha, I was missing that implicit relationship. Thanks for the rewrite though, that at least makes sure it behaves correctly, even if the context doesn't get torn down like it should.

wbobeirne (approved):

Issue #12, opened 2023-05-08
Don't hardcode modules in SetConfiguration.tsx


garyray-k:
See TODO in `SetConfiguration.tsx`

_Originally posted by @garyray-k in https://github.com/fedimint/fedimint/pull/2360#discussion_r1186956451_

justinmoon:
I think the big thing is that we shouldn't throw away anything in we receive in `get_default_config_gen_params`. Doing so is what caused the regression fixed by https://github.com/fedimint/ui/pull/106. Our current strategy doesn't notice new defaults.

justinmoon:
Closed by https://github.com/fedimint/ui/pull/133

PR #106, opened 2023-07-21
fix: include default client bitcoin rpc config in params

justinmoon:
This block of code handling the config gen params is very error-prone. Related to https://github.com/fedimint/ui/issues/12.

### DIFF
--- a/apps/guardian-ui/src/components/SetConfiguration.tsx
+++ b/apps/guardian-ui/src/components/SetConfiguration.tsx
@@ -90,10 +90,9 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
         );
         setNetwork(walletModule.consensus?.network.toString() || network);
         setBitcoinRpc(walletModule.local?.bitcoin_rpc || bitcoinRpc);
-        setClientDefaultBitcoinRpc(
-          walletModule.consensus?.client_default_bitcoin_rpc ||
-            clientDefaultBitcoinRpc
-        );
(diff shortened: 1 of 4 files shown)


apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -85,6 +90,10 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
         );
         setNetwork(walletModule.consensus?.network.toString() || network);
         setBitcoinRpc(walletModule.local?.bitcoin_rpc || bitcoinRpc);
+        setClientDefaultBitcoinRpc(
+          walletModule.consensus?.client_default_bitcoin_rpc ||
+            clientDefaultBitcoinRpc
+        );
```
wbobeirne, line 96: Hmmm this is a bit of an anti-pattern but not your fault. But these are using variables that are not being included in the `useEffect` dependencies, so they may be stale references. Typically you'd just wrap this in a condition and not call the `set*` method, rather than re-passing its existing state:
```suggestion
        if (walletModule.consensus?.client_default_bitcoin_rpc) {
          setClientDefaultBitcoinRpc(walletModule.consensus.client_default_bitcoin_rpc);
        }
```
However, `walletModule` isn't in the `useEffect` dependencies either so it's kind of a wash. This whole `useEffect` probably needs to be re-thought.

wbobeirne (approved):

PR #116, opened 2023-07-24
Wo/lint hooks

### LINKED ISSUE #107: Add `eslint-plugin-react-hooks`
We've got quite a few places in the codebase where `useEffect`s don't have dependencies properly setup, and we have stale references. We should add https://www.npmjs.com/package/eslint-plugin-react-hooks to the eslint config.

wbobeirne:
Closes #107

### What This Does

* Installs `eslint-plugin-react-hooks` and configures it in `@fedimint/eslint-config`
* Fixes all instances of missing or unnecessary hook dependencies
  * These will likely fix any subtle bugs that may have occurred if a hook had stale references to old state
* Minor logic changes to reduce hook dependencies / reduce unnecessary re-runs of hooks


apps/guardian-ui/src/AppContext.tsx
```diff
@@ -31,7 +33,6 @@ export interface AppContextProviderProps {
 export const AppContextProvider: React.FC<AppContextProviderProps> = ({
   children,
 }: AppContextProviderProps) => {
-  const api = new GuardianApi();
```
wbobeirne, line 34: This was causing every re-render of this context to create a new API instance. Bad times!

packages/eslint-config/index.js
```diff
@@ -1,24 +1,29 @@
...
   extends: [
-    "eslint:recommended",
-    "plugin:@typescript-eslint/recommended",
-    "plugin:react/recommended",
-    "turbo",
-    "prettier",
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react/recommended',
+    'plugin:react-hooks/recommended',
```
wbobeirne, line 6: Not sure why this package wasn't prettier'd before, but it looks like my IDE picked up the prettier config and formatted this. Adding this line was the only change.

apps/guardian-ui/src/AppContext.tsx
okjodom: cool memo

okjodom (approved):
LGTM! Lots of improvements

apps/guardian-ui/src/components/VerifyGuardians.tsx
okjodom, line 1: Change introduces a bug in `VerifyGuardians`, where the first time (s) we poll consensus state, we find the websocket closed. Narrowed down to second commit, but still debugging

okjodom: Fixed. Bug was in a `AppContext`, where we would still shutdown api as a side effect of context dismount. I needed to appropriately resolve this with the last commit which removes similar logic previously in `SetupContext`

apps/guardian-ui/src/App.tsx
```diff
@@ -15,7 +15,7 @@ export const App = React.memo(function App() {
   const { t } = useTranslation();
   const { state, api } = useAppContext();
 
-  const getAppContent = useCallback(() => {
```
EthnTuttle, line 18: TIL: https://medium.com/@jan.hesters/usecallback-vs-usememo-c23ad1dc60

EthnTuttle (approved):

PR #125, opened 2023-07-26
feat: use git tags for fedimint docker images

justinmoon:
Using `master` tag is error-prone because if you've already pulled it and an update is made to backend which the UI needs, docker won't re-download this new tag unless you manually run a command like `docker pull fedimintd` / `docker pull gatewayd`.


okjodom (approved):

IgboPharaoh:
This actually fixed my problem.

PR #132, opened 2023-08-01
Refactor/gateway admin

IgboPharaoh:
This is a cleanup pr for gateway-admin. Since we're using cards instead of tabs, it makes sense to rename Components and language typings to match the design style


okjodom (approved):
appreciating the code cleanups

PR #133, opened 2023-08-01
Don't throw away 3rd party modules

justinmoon:
This allows the UI to run 3rd party modules. It doesn't expose their parameters in the UI so you can only run with default parameters. But it's a good step forward. I was able to test it with a 3rd party module and it worked.

@wbobeirne could use a little help with typing things.

Updated docker and nix fedimint hashes to include [this commit](https://github.com/fedimint/fedimint/pull/2848) which enables devimint to do DKG with 3rd party modules.

### DIFF
--- a/apps/guardian-ui/src/components/SetConfiguration.tsx
+++ b/apps/guardian-ui/src/components/SetConfiguration.tsx
@@ -20,12 +20,10 @@ import {
 import { FormGroup, FormGroupHeading } from '@fedimint/ui';
 import { useSetupContext } from '../hooks';
 import {
-  AnyModuleParams,
   BitcoinRpc,
   ConfigGenParams,
   GuardianRole,
   Network,
-  OtherModuleParams,
 } from '../setup/types';
 import { ReactComponent as FedimintLogo } from '../assets/svgs/fedimint.svg';
 import { ReactComponent as BitcoinLogo } from '../assets/svgs/bitcoin.svg';
@@ -33,7 +31,8 @@ import { ReactComponent as ArrowRightIcon } from '../assets/svgs/arrow-right.svg
 import {
   formatApiErrorMessage,
   getModuleParamsFromConfig,
-  getOtherModuleParamsFromConfig,
+  applyConfigGenModuleParams,
+  removeConfigGenModuleConsensusParams,
 } from '../utils/api';
 import { ModuleKind } from '../types';
 import { useTranslation } from '@fedimint/utils';
@@ -60,6 +59,7 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
   const [myName, setMyName] = useState(stateMyName);
   const [password, setPassword] = useState(statePassword);
   const [hostServerUrl, setHostServerUrl] = useState('');
+  const [defaultParams, setDefaultParams] = useState<ConfigGenParams>();
   const [numPeers, setNumPeers] = useState(
     stateNumPeers ? stateNumPeers.toString() : ''
   );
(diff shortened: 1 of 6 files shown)


apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -142,33 +150,36 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
...
-                'ln',
-                {
-                  consensus: { network: network as Network },
-                  local: { bitcoin_rpc: bitcoinRpc },
-                },
-              ],
-            },
+                ],
+              },
+              otherModuleParams as Record<number, AnyModuleParams> // FIXME
```
justinmoon, line 181: This is a hack

apps/guardian-ui/src/utils/api.ts
```diff
@@ -15,6 +16,26 @@ export function getModuleParamsFromConfig<T extends AnyModuleParams[0]>(
...
+      (key) =>
+        config.modules[parseInt(key)][0] != ModuleKind.Mint &&
+        config.modules[parseInt(key)][0] !== ModuleKind.Ln &&
+        config.modules[parseInt(key)][0] !== ModuleKind.Wallet
+    )
+    .reduce((cur, key) => {
+      return Object.assign(cur, { [key]: config.modules[parseInt(key)] });
+    }, {});
+}
+
```
justinmoon, line 38: Notice the `parseInt` calls here. Seems like our types are kinda wrong? I would expect to be able to just index into an object with whatever keys it has.

wbobeirne: Interfaces in TypeScript can be subsets of the actual underlying object, therefore any call to `Object.keys` or other key-grabbing functions will always loosen the type to `string`.

I'll have to check the way the API serializes it, but I think I originally saw the keys being numeric and typed them to number, but maybe we should just treat them like strings since object keys are always strings.

apps/guardian-ui/src/components/SetConfiguration.tsx
wbobeirne: I don't think this fully soles the issue because `Object.assign` is not a deep assignment, meaning if `otherModuleParams` included a module param like `client_default_bitcoin_rpc` where we were defining some of the wallet params, but not all of them, it would get blown out. I think we need to do a deep merge of these objects instead.

apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -71,10 +77,12 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
     });
   const [mintAmounts, setMintAmounts] = useState<number[]>([]);
   const [error, setError] = useState<string>();
+  const [otherModuleParams, setOtherModuleParams] = useState<object>({});
 
   useEffect(() => {
     const initStateFromParams = (params: ConfigGenParams) => {
       setFederationName(params.meta?.federation_name || '');
+      setOtherModuleParams(getOtherModuleParamsFromConfig(params));
```
wbobeirne, line 85: IMO this is not necessary to filter out the old modules and makes this unnecessarily complicate. Just store `params`, and then we can call `submitConfiguration` with:
```ts
submitConfiguration({
  modules: {
    ...params,
    // all the known 1st party modules after the spread
  },
);
```

wbobeirne (changes requested):
I think we could fix this up for the 3rd party module case, but we're still leaving two important issues in here:

1. Hard coding of 1st party module IDs, when that's not guaranteed
2. Omitting 1st party module keys the frontend doesn't know about

I don't want to let great be the enemy of good, but I kind of think we should solve those holistically since solving those will probably end up throwing this code out.

wbobeirne (approved):

justinmoon:
Nice. Now we're not hard-coding instance IDs which is a big improvement on top of being able to run external modules.

PR #142, opened 2023-08-16
fix: no verifyGuardians page for solo + JUSTFILE

Kodylow:
- adds a catch to skip showing verifyguardians if it's a 1/1 federation.
- adds a JUSTFILE for commands

### DIFF
--- a/apps/guardian-ui/src/components/VerifyGuardians.tsx
+++ b/apps/guardian-ui/src/components/VerifyGuardians.tsx
@@ -86,6 +86,11 @@ export const VerifyGuardians: React.FC<Props> = ({ next }) => {
   }, [api]);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


EthnTuttle (approved):

apps/guardian-ui/src/components/VerifyGuardians.tsx
```diff
@@ -86,6 +86,11 @@ export const VerifyGuardians: React.FC<Props> = ({ next }) => {
   }, [api]);
 
   useEffect(() => {
+    // If we're the only guardian, skip this verify other guardians step.
+    if (numPeers === 1) {
+      next();
```
EthnTuttle, line 91: Need to `await api.startConsensus();` here for we `next()`

PR #144, opened 2023-08-18
Colorize online status

EthnTuttle:
Begin work on some colorization

### DIFF
--- a/apps/guardian-ui/src/admin/FederationAdmin.tsx
+++ b/apps/guardian-ui/src/admin/FederationAdmin.tsx
@@ -33,6 +33,9 @@ export const FederationAdmin: React.FC = () => {
   const [config, setConfig] = useState<ConfigResponse>();
   const [gateways, setGateways] = useState<Gateway[]>([]);
   const [guardians, setGuardians] = useState<string | undefined>();
+  const [statusColor, setStatusColor] = useState<
+    'red' | 'green' | 'yellow' | undefined
+  >();
   const { t } = useTranslation();
 
   useEffect(() => {
@@ -69,6 +72,14 @@ export const FederationAdmin: React.FC = () => {
     const offline = status?.federation ? status.federation.peers_offline : 0;
 
     const totalPeers = online + offline;
+    const onlinePercentage = online / totalPeers;
+    if (onlinePercentage === 1) {
+      setStatusColor('green');
+    } else if (onlinePercentage >= 2 / 3) {
+      setStatusColor('yellow');
+    } else {
+      setStatusColor('red');
+    }
     setGuardians(`${online} / ${totalPeers}`);
   }, [status]);
 
@@ -98,7 +109,11 @@ export const FederationAdmin: React.FC = () => {
               {t('federation-dashboard.placeholder-fed-description')}
             </Text>
             <Flex gap='12px' mt='13px'>
-              <Pill text='Guardians' status={`${guardians}`} />
+              <Pill
+                text='Guardians'
+                status={`${guardians}`}
+                color={statusColor}
+              />
               <Pill text='Server' status='Healthy' />
               <Pill text='Uptime' status='100%' />
             </Flex>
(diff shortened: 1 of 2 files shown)


apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -69,6 +72,14 @@ export const FederationAdmin: React.FC = () => {
...
 
     const totalPeers = online + offline;
+    const onlinePercentage = online / totalPeers;
+    if (onlinePercentage === 1) {
+      setStatusColor('green');
+    } else if (onlinePercentage >= 2 / 3) {
+      setStatusColor('yellow');
+    } else {
+      setStatusColor('red');
+    }
```
wbobeirne, line 82: All of the `set*` calls in here should just be calculated in the render, not in the `useEffect`. That causes an unnecessary second render. It goes

1. `status` reference changes
2. render runs (with old `guardians` and `statusColor`)
3. `useEffect` runs where `set*` calls are made
4. render runs again (with updated values)

Since these are simple values where we don't need to maintain a reference, they can just be calculated and assigned each render.

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -98,7 +109,11 @@ export const FederationAdmin: React.FC = () => {
...
             </Text>
             <Flex gap='12px' mt='13px'>
-              <Pill text='Guardians' status={`${guardians}`} />
+              <Pill
+                text='Guardians'
+                status={`${guardians}`}
+                color={statusColor}
+              />
               <Pill text='Server' status='Healthy' />
               <Pill text='Uptime' status='100%' />
```
wbobeirne, line 118: Do these need to have `color="green"` added?

apps/guardian-ui/src/admin/FederationAdmin.tsx
EthnTuttle: I appreciate your React framework feedback.

apps/guardian-ui/src/admin/FederationAdmin.tsx
EthnTuttle: Good call. They're static values so I'd love to remove them but 🤷‍♂️. 
https://github.com/fedimint/ui/issues/148 for tracking. Need to please bring this up in the dev call.

CC: @justinmoon in case I can't make the call.

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -67,21 +67,19 @@ export const FederationAdmin: React.FC = () => {
     }
   }, [config, api]);
 
-  useEffect(() => {
-    const online = status?.federation ? status.federation.peers_online + 1 : 1;
-    const offline = status?.federation ? status.federation.peers_offline : 0;
+  const online = status?.federation ? status.federation.peers_online + 1 : 1;
```
okjodom, line 70: React flags [this commit](https://github.com/fedimint/ui/pull/144/commits/45a6e92c7881fd4d8f272f12ea20257252032c0d) as causing multiple (infinite) rerenders

okjodom: Probably need to calculate the pill props under a memo, removing `gaurdians` and `statusColor` from state

```ts
const [guardians, statusColor] = useMemo(() => {
    const online = status?.federation ? status.federation.peers_online + 1 : 1;
    const offline = status?.federation ? status.federation.peers_offline : 0;
    const totalPeers = online + offline;

    const onlinePercentage = online / totalPeers;
    const guardans = `${online} / ${totalPeers}`;

    const statusColor: 'green' | 'yellow' | 'red' =
      onlinePercentage === 1
        ? 'green'
        : onlinePercentage >= 2 / 3
        ? 'yellow'
        : 'red';

    return [guardans, statusColor];
  }, [status]);
  ```
  
  `"guardians"` string could use a better variable name?

EthnTuttle: I appreciate the feedback and patience. I have implemented the useMemo solution and did some reading on the topic.

EthnTuttle: @okjodom Where did React tell you about the infinite re-render? So I can pay better attention to that moving forward.

okjodom: Saw it in uncaught runtime errors
image

okjodom (approved):

Issue #122, opened 2023-07-26
Leader shows Pending for self


EthnTuttle:
On the Invite Guardians page, the Leader shows a "Pending" state, although they are Ready. This may be confusing as it suggests the Leader still has action to take.

This label should be marked differently to be less confusing.

EthnTuttle:
Closed by #146

PR #146, opened 2023-08-18
fix: use correct conditional for status

EthnTuttle:


EthnTuttle:
Closes #122

wbobeirne (approved):
My mistake for approving the last PR without thoroughly checking, ideally this would actually check the role of the peer but incidentally I think this works because this table only renders once you have approved, but I could see this being error-prone if we ever decided to render the peer table before that. But for now this fixes the incorrect implementation 👌

PR #150, opened 2023-08-20
build: four fedimintds for dev env

EthnTuttle:
I used four fedimints since I imagine this being the "standard". We can now also dev on UI/UX for when consensus concerns arise by turning off a fedimint or two.

Marking Draft because I haven't actually tested a setup with this, but wanted to get it out there.

### DIFF
--- a/mprocs.yml
+++ b/mprocs.yml
@@ -19,6 +19,18 @@ procs:
       PORT: '3001'
       REACT_APP_FM_CONFIG_API: ws://127.0.0.1:18184
       BROWSER: none
+  guardian-ui-3:
+    shell: yarn dev:guardian-ui
+    env:
+      PORT: '3003'
+      REACT_APP_FM_CONFIG_API: ws://127.0.0.1:18185
+      BROWSER: none
+  guardian-ui-4:
+    shell: yarn dev:guardian-ui
+    env:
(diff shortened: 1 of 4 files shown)


EthnTuttle:
Tested. Works.

mprocs.yml
```diff
@@ -19,6 +19,18 @@ procs:
       PORT: '3001'
       REACT_APP_FM_CONFIG_API: ws://127.0.0.1:18184
       BROWSER: none
+  guardian-ui-3:
+    shell: yarn dev:guardian-ui
+    env:
+      PORT: '3003'
```
okjodom, line 25: nit: maybe we need to assign contiguous ports to the guardian UIs thus push gateway UI port back to 3004 or 3005?

okjodom (dismissed):
validated. left a nit/suggestion about gateway ui port

okjodom:
nit: should also add all `fm_*` dirs to gitignore

mprocs.yml
EthnTuttle: Good call. Was lazy. 😅

EthnTuttle:
@okjodom Addressed the nits. Good call so we can keep this repo well "gardened".

okjodom (approved):

PR #151, opened 2023-08-20
94 guardian show list of guardians in federation

EthnTuttle:
Displays Peer Id and Name alongside connection status.
Fixes typo in API endpoint field name.

### DIFF
--- a/apps/guardian-ui/src/admin/FederationAdmin.tsx
+++ b/apps/guardian-ui/src/admin/FederationAdmin.tsx
@@ -17,12 +17,47 @@ import {
 import { CopyInput } from '@fedimint/ui';
 import { useTranslation } from '@fedimint/utils';
 import { useAdminContext } from '../hooks';
-import { ConfigResponse, Gateway, StatusResponse, Versions } from '../types';
+import {
+  ClientConfig,
+  ConfigResponse,
+  FederationStatus,
+  Gateway,
+  PeerStatus,
+  StatusResponse,
+  Versions,
+} from '../types';
 import { AdminMain } from '../components/AdminMain';
 import { ConnectedNodes } from '../components/ConnectedNodes';
 import { LightningModuleRpc } from '../GuardianApi';
 import { ReactComponent as CopyIcon } from '../assets/svgs/copy.svg';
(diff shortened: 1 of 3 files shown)


apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -17,13 +17,48 @@ import {
...
 import { ReactComponent as CopyIcon } from '../assets/svgs/copy.svg';
 import { Pill } from '../components/Pill';
 
+interface PeerData {
+  id: number;
+  name: string;
+  status: PeerStatus;
+}
+
+function generatePeerData(
```
okjodom, line 41: style: I'd just inline this function within the useEffect. It's not reused

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -17,13 +17,48 @@ import {
...
+  config: ClientConfig,
+  federationStatus: FederationStatus
+): PeerData[] {
+  const peerDataArray: PeerData[] = [];
+  for (const [id, status] of Object.entries(federationStatus.status_by_peer)) {
+    const numericId = parseInt(id, 10);
+    if (Object.prototype.hasOwnProperty.call(config.api_endpoints, numericId)) {
```
okjodom, line 48: can we skip the outer `hasOwnProperty` check? It doesn't seem very useful if endpoint is still nullable and we can just go with 
```ts
const endpoint = config.api_endpoints[numericId];
      if (endpoint) {
        peerDataArray.push({
          id: numericId,
          name: endpoint.name,
          status,
        });
      }
      ```

okjodom:
needs rebase after #144 ?

EthnTuttle:
> needs rebase after #144 ?

I don't believe so since this works on a different component. #144 was working on the header status bar and this is the lower right card with Peer status. 
We should probably colorize the Peers in some way too?

wbobeirne (approved):
Worked great 👍 

Screenshot 2023-08-29 at 5 13 43 PM

wbobeirne:
It did require a rebase though, I rebased and resolved conflicts.

PR #152, opened 2023-08-20
feat: QR for fed1 invite code

EthnTuttle:
image

image

### DIFF
--- a/apps/guardian-ui/package.json
+++ b/apps/guardian-ui/package.json
@@ -23,10 +23,13 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-icons": "^4.7.1",
+    "react-qr-code": "^2.0.12",
     "react-scripts": "5.0.1",
     "react-use-websocket": "3.0.0"
   },
   "devDependencies": {
+    "@fedimint/eslint-config": "*",
+    "@fedimint/tsconfig": "*",
     "@testing-library/jest-dom": "^5.16.1",
     "@testing-library/react": "^13.0.0",
     "@testing-library/user-event": "^13.5.0",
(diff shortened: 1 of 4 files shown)


apps/guardian-ui/package.json
```diff
@@ -23,10 +23,13 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-icons": "^4.7.1",
+    "react-qr-code": "^2.0.12",
```
wbobeirne, line 26: This seems like a decently popular library, although relies it on `qr.js` which appears to no longer be maintained (last update 10 years ago) and is hard to track down the source code for. But swapping QR implementations is pretty trivial, so I'm not worried about what's chosen as long as it doesn't look like a security risk.

apps/guardian-ui/src/languages/en.json
```diff
@@ -17,7 +17,7 @@
   "federation-dashboard": {
     "placeholder-fed-description": "A community-based coalition focused on ushering the world into a bitcoin standard",
     "invite-members": "Invite members",
-    "invite-members-prompt": "Share this to invite members to join the federation",
+    "invite-members-prompt": "Share this to invite members to join the federation. Or have them scan the QR code ->",
```
wbobeirne, line 20: One thing to consider with the `->` is that with text wrapping / layout changes in different languages or smaller screen widths, the QR code may not always be directly to the right of this copy. As @Kodylow noted in Discord, it might be good to have it very clear what this QR code is for inline.

wbobeirne (commented):
Personally I think this is a net improvement even if we want the design community to do another pass to it. I'll hold off on hitting "Approve" so that some others can take a look before merging.

I think in general we should avoid deviating from the Figma _too_ much so that changes to the interface have a process to go through rather than just being at the whims of whoever last worked on a UI element, but I think this is an improvement and a good starting point for design to incorporate into figma going forward.

EthnTuttle:
Marking draft until design can review.

EthnTuttle:
Closing for now due to conflicts. Not too intricate that it can't be replicated easily again.

PR #158, opened 2023-08-28
fix: update fedimint version to v0.1 branch

justinmoon:
Use v0.1.0-rc3 git tag for docker and nix fedimint deps.

I can complete the setup using docker, but sometimes need to refresh the last page to get to the dashboard. Not sure what is causing that.

Also, the docker setup doesn't mine any blocks (this is one of many nice things that devimint automates for us) and so it seems like after DKG completes the federation just keeps producing tons of rounds of consensus trying to get a first bitcoin block.


wbobeirne:
Don't forgot the image tags in `docker-compose.yml`

wbobeirne (approved):

PR #164, opened 2023-08-30
Clean up theme colors & component hex values

wbobeirne:
Untracked cleanup

### What This Does

* Adds `red`, `yellow`, and `green` colors to palette
  * I've aliased these as `success`, `warning` and `error` to match the figma but it's pretty common for themes to both have named colors and aliased colors so that, for instance, you could have a theme that uses a different palette color for a state, e.g. a purple warning or a blue success. I think we should do this with all of the colors like `blue` and `gray` at some point with aliases like `primary` and `brand`. We already somewhat do this with `text` and `border`.
* Updates all components to use `theme.colors.*` instead of hard-coded hex values
  * Tweaked `Pill` component to better match designs overall
    * Renamed `color` prop to `type` so that it matches `success | warning | error` nomenclature
* Removes unused components
  * `guardian-ui`'s `<SideBar />`
  * `gateway-ui`'s `<Input />`
  * I removed them as a part of this PR because I wanted to get rid of all hex values, but had no way to visually confirm changes to these since there is currently no way to get these to render. In general I think we should not leave dead code in the project as we can always restore them via git later.

## Screenshots

Screenshot 2023-08-30 at 1 42 43 PM
Screenshot 2023-08-30 at 1 40 58 PM
Screenshot 2023-08-30 at 1 43 09 PM
Screenshot 2023-08-30 at 1 43 20 PM


EthnTuttle (approved):
pulled and tested. 

hex colors are dead, long live the theme!

PR #168, opened 2023-08-31
Update Guardian Admin UI to match Figma spec

### LINKED ISSUE #27: CopyInput doesn't allow viewing hyper-long strings
connectionCode in apps/guardian-ui/src/components/FederationDashboard.tsx is an example. The connection info string is very long cannot be viewed in full in the CopyInput element.

### LINKED ISSUE #64: Guardian: Show information about connected lightning nodes / gateway
For each lightning node connected to the federation, we should show public node and routing information

- [x] Fetch list of connected gateways using guardian api : https://github.com/fedimint/ui/pull/69
- [ ] Update lightning node info UX and show info
   - title gateways using gateway id
   - view gateway node using node id

image

### LINKED ISSUE #95: Guardian: Show federation info
Polish UI to show these Federation info
- [ ] Consensus info
- [ ] Bitcoin balance
- [ ] Bitcoin node info

image

### LINKED ISSUE #148: FederationAdmin.tsx Dashboard Static Server Health
The indicators for the Guardians own server just have static values. We need to hook up to an API (maybe needs backend work?) and determine what metric we're using for this information.

### LINKED ISSUE #166: Update guardian UI admin to match Figma spec designs
Figma spec: https://www.figma.com/file/ioeAVbWBc3ow2RaZETatM7/Fedimint?type=design&node-id=5520-77905&mode=design&t=7oo4vc9U0OeSrnuS-0

Current implementation on the left vs figma spec on the right

Screenshot 2023-08-30 at 2 48 50 PM

Screenshot 2023-08-30 at 2 48 38 PM

Main differences that need updating is:

* [ ] No description text below federation name
* [ ] Remove pills from the top
* [ ] All sections should use thick bottom border cards
* [ ] "Federation Info" should be a list instead of a table
* [ ] Add a "Bitcoin Node" section
* [ ] Single card for balance instead of "Community Custody" section
* [ ] "Peer Info" to become "Guardians" with designed Table component
* [ ] "Gateway" to become "Lightning Gateways" using that same Table component

Note that this issue does not intend to properly hook up this information to the API, merely to get the designs to match. Data hook-ups are their own issues that can be handled independently (#124, #95, #64)

wbobeirne:
Closes #166, closes #64, closes #95 (sorta?), closes #27, closes #148

Apologies for the size of this PR. Doing it all at once made it easier to get the organization and component abstractions right the first time, rather than having to re-work it a piece at a time.

### What This Does
* Updates shared UI components to match figma spec and adds new reusable components
  * Chakra's `Card` component now matches Figma spec
  * Custom `Table` component now matches Figma spec
  * Custom `CopyInput` component now matches Figma spec
    * Added new `sm` size support
    * `<Input readOnly />` now renders with lighter text
  * Custom `Wrapper` component now matches Figma spec
    * Added new `size="md|lg"` prop for admin's wider 1200px template
  * Added new `StatusIndicator` component
    * This used to be `Pill` in `guardian-ui`, but no longer comes with the label prefix
  * Added new `KeyValues` component for displaying lists of information with a label on top and a value below
* Updated guardian UI admin to match Figma spec
  * See commit history and checklist in #166 for full list of changes
  * Some components are still stubbed out due to missing data, and have other issues open to address them
* Lightly tweaked gateway UI to be closer to Figma spec
  * Ellipsized node id
  * `<Wrapper size="lg">` to match Figma spec width
  * Refactored to use new `Card` styles
  * Gateway UI is still not 1:1 with Figma spec

Note that some deviations to the Figma spec were made to match the information we have available:
* Federation Info's "Consensus version" is a single digit, not a major.minor
* Bitcoin Node just shows "Unknown" until we get a way to fetch non-encoded module configs
* Guardian table's "Last contribution timestamp" was replaced with "Last contribution epoch"
* Gateway table's "Name" was replaced with "Node ID" due to not having the alias
* Gateway table has an additional "Gateway ID" column
* Gateway table's "Incoming fee" and "Outgoing fee" were replaced with a single "Gateway fee"

### Guardian UI admin

Screenshot 2023-08-31 at 11 58 59 AM

### Gateway UI

Screenshot 2023-08-31 at 12 16 25 PM


packages/ui/src/Wrapper.tsx
```diff
@@ -18,7 +20,7 @@ export const Wrapper = memo(function Wrapper({
       alignItems='center'
     >
       <Box
-        maxW='960px'
+        maxW={size === 'lg' ? '1200px' : '960px'}
```
EthnTuttle, line 23: So `md` is just 960px?

packages/ui/src/CopyInput.tsx
```diff
@@ -21,26 +23,43 @@ export const CopyInput: React.FC<CopyInputProps> = ({
 }) => {
   const { onCopy, hasCopied } = useClipboard(value);
   const theme = useTheme();
+  const [buttonWidth, buttonHeight] = {
```
EthnTuttle, line 26: Where do we derive these ratios from?
I noticed in the admin dashboard, that the button overlaps and covers the input border and looks weird. When I adjusted `sm` to have a height of `28px`, the issue went away.
Before:

Screenshot 2023-09-01 at 11 29 14 AM

After (`28px` for `sm`):

Screenshot 2023-09-01 at 11 31 44 AM

apps/guardian-ui/src/languages/en.json
```diff
@@ -1,8 +1,10 @@
 {
   "common": {
+    "bitcoin": "Bitcoin",
```
EthnTuttle, line 3: nit: when used as a measurement or unit, it would be `bitcoin`. `Bitcoin` with a capital B is for the network or protocol. (I'll take "things that don't matter that much to anyone else" for 500, Alex)

apps/guardian-ui/src/components/GuardiansCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+      },
+      {
+        key: 'lastContribution',
+        heading: t('federation-dashboard.guardians.last-contribution-label'),
+      },
+    ],
+    [t]
+  );
+
+  const rows: TableRow<TableKey>[] = useMemo(() => {
```
EthnTuttle, line 39: We should still surface "Peer ID" as a value to guardians since this will be how their fedimints are referenced in logs. If I'm a Guardians troubleshooting logs, I don't want the mental burden of remembering if Alice or Bob were Peer 0 or Peer 1.

apps/gateway-ui/src/components/DepositCard.tsx
```diff
@@ -44,95 +44,81 @@ export const DepositCard = React.memo(function DepositCard({
...
+        </Box>
       </Stack>
+      <Flex gap='4px'>
+        <Link
+          fontSize='md'
+          color={theme.colors.blue[600]}
+          fontFamily={theme.fonts.body}
+          _hover={{ textDecoration: 'underline' }}
+          transition={`text-decoration 1s ease-in-out`}
+          href={`https://mempool.space/address/${address}`}
```
EthnTuttle, line 113: Random thought: Settings to configure what mempool you want to use or LN explorer, etc. Def future fedimint problems to address.

apps/gateway-ui/src/components/WithdrawCard.tsx
```diff
@@ -96,35 +96,16 @@ export const WithdrawCard = React.memo(function WithdrawCard({
...
-          <Text
-            cursor='pointer'
-            fontSize='sm'
-            color={theme.colors.blue[600]}
-            fontFamily={theme.fonts.body}
-          >
-            {t('withdraw-card.withdraw_all')}
-          </Text>
-        </Stack>
+    <>
```
EthnTuttle, line 104: I don't see this pattern in the other cards?

apps/gateway-ui/src/components/WithdrawCard.tsx
```diff
@@ -96,35 +96,16 @@ export const WithdrawCard = React.memo(function WithdrawCard({
       });
   };
 
+  const description = `${t('withdraw-card.total_bitcoin')} ${
+    withdrawObject.amount / 100000
```
EthnTuttle, line 100: Is this converting milli-sats to bits? Should we divide by 1000 to have just sats? OMG ARE YOU TEAM BITS?!?! 🤣

apps/gateway-ui/src/components/WithdrawCard.tsx
EthnTuttle, line 1: Just want to leave a general note that we should make some of these existing comments into Issues if they're not.

EthnTuttle (commented):
Not finished but left some comments.

Still need to review:
- apps/guardian-ui/src/admin/FederationAdmin.tsx
- apps/guardian-ui/src/components/BalanceCard.tsx
- apps/guardian-ui/src/components/BitcoinNodeCard.tsx
- apps/guardian-ui/src/components/FederationInfoCard.tsx
- apps/guardian-ui/src/components/GatewaysCard.tsx

packages/ui/src/Wrapper.tsx
wbobeirne: Yep, or more literally, anything not `lg` is `960px`. But the prop is typed as `'lg' | 'md'` so in practice that means `md` is `960px`

packages/ui/src/CopyInput.tsx
wbobeirne: What browser is this? It's possible there's some rendering quirks due to how precisely this is placed. Your `28px` has these white gaps around the button that we want to avoid. The goal of these heights is to make it fit within the input's borders, so I think they all net out to `inputHeight - 2px`. However I don't love the way these are done, but it's to deal with Chakra's lack of built-in support for a button inside of an input that's flush with the borders.

If we're willing to deviate from the design, we could place the button flush _beside_ the input, or we can place the button with some padding inset inside the input: https://chakra-ui.com/docs/components/input.

apps/guardian-ui/src/components/GuardiansCard.tsx
wbobeirne: Agreed, I'll add another column and a note about deviating from designs, similar to how I did with `Gateway ID`

apps/gateway-ui/src/components/WithdrawCard.tsx
wbobeirne: This is due to the `<ConfirmWithdrawModal />` at the bottom. React components can only render a single node, which is why this was previously wrapped in a `<Box />`. However, that adds an unnecessary extra `<div>` to the markup. Using a `<>` React Fragment means we return one React node, but it doesn't render excess markup in the DOM that can effect layout.

apps/gateway-ui/src/components/DepositCard.tsx
wbobeirne: Agreed. It's _slightly_ less of a concern since we're not using these as APIs, just linking out, but configurability is king.

apps/gateway-ui/src/components/WithdrawCard.tsx
wbobeirne: I actually didn't write this code, I just pulled it out of the render into a var but it was originally written here: https://github.com/fedimint/ui/blob/dca2a26bb8817c075756e74354cd907ff4b46f23/apps/gateway-ui/src/components/WithdrawCard.tsx#L116. I haven't actually checked if the math is valid, but this should probably use a utility function instead of a one-off implementation due to how easy it is to be off by an order of magnitude or two.

apps/gateway-ui/src/components/WithdrawCard.tsx
wbobeirne: Agreed, we should probably do a `TODO` audit of the ui codebase since I think there are a _lot_ of these sprinkled in. `git grep -l TODO | xargs -n1 git blame | grep TODO` is a handy macro.

packages/ui/src/CopyInput.tsx
wbobeirne: I've adjusted the code here to be more dynamic, and was unable to replicate your issue in any browser:

Screenshot 2023-09-01 at 5 46 52 PM

LMK if you still see it, maybe you can tinker with the inspector to figure out what's wrong?

packages/ui/src/CopyInput.tsx
```diff
@@ -21,26 +22,48 @@ export const CopyInput: React.FC<CopyInputProps> = ({
...
+              {buttonLeftIcon}
+            </Flex>
+          }
           onClick={onCopy}
+          borderRadius={theme.components.Input.sizes[size].borderRadius}
           borderTopLeftRadius={0}
           borderBottomLeftRadius={0}
           size={size}
-          height={size == 'lg' ? '46px' : '42px'}
+          height='100%'
```
EthnTuttle, line 62: Shouldn't this be `buttonHeight`?
```suggestion
          height={buttonHeight}
```

EthnTuttle (changes requested):

EthnTuttle:
Going to approve and merge to unlock all the other PRs.

EthnTuttle (approved):

packages/ui/src/CopyInput.tsx
EthnTuttle: I no longer see it. Maybe just a glitch in my env.

apps/gateway-ui/src/components/DepositCard.tsx
EthnTuttle: #181

apps/gateway-ui/src/components/WithdrawCard.tsx
EthnTuttle: https://github.com/fedimint/ui/issues/182

apps/guardian-ui/src/components/GuardiansCard.tsx
EthnTuttle: Design is updating Figma (see Discord)

PR #170, opened 2023-08-31
Hook up `audit` rpc to UI balance

wbobeirne:
Built on #168, fixes part of #95

Shout out @EthnTuttle for getting started on this

### What This Does

* Adds the `api.audit` call for federation & module assets and liabilities
* Hooks up the `wallet` module's `net_assets` to the `<BalanceCard />`
* Adds new shared types to `packages/utils`
  * Opaque number types have been added for `Btc`, `Sats`, and `MSats` that makes it so that API types and functions can dictate which specific units they accept, to avoid the ambiguity of `number` not being clear what denomination it is
* Adds missing `docker-compose.yml` `depends_on: bitcoin` configs for fedimintds

## Screenshots

Screenshot 2023-08-31 at 4 31 14 PM

### DIFF
--- a/apps/guardian-ui/src/components/BalanceCard.tsx
+++ b/apps/guardian-ui/src/components/BalanceCard.tsx
@@ -1,21 +1,37 @@
-import { Card, CardBody, CardHeader, Text } from '@chakra-ui/react';
+import { Card, CardBody, CardHeader, Skeleton, Text } from '@chakra-ui/react';
 import { KeyValues } from '@fedimint/ui';
-import { useTranslation } from '@fedimint/utils';
-import React, { useMemo } from 'react';
+import { MSats, formatMsatsToBtc, useTranslation } from '@fedimint/utils';
+import React, { useEffect, useMemo, useState } from 'react';
+import { AuditSummary } from '../types';
+import { useAdminContext } from '../hooks';
 
 export const BalanceCard: React.FC = () => {
   const { t } = useTranslation();
+  const { api } = useAdminContext();
+  const [auditSummary, setAuditSummary] = useState<AuditSummary>();
+
+  const walletBalance = auditSummary
+    ? auditSummary.module_summaries.wallet?.net_assets || (0 as MSats)
+    : undefined;
+
+  useEffect(() => {
+    api.audit().then(setAuditSummary).catch(console.error);
+  }, [api]);
 
   const keyValues = useMemo(
     () => [
       {
         key: 'bitcoin',
         label: t('common.bitcoin'),
-        // TODO: Use `audit` api to fill this in
-        value: '0.00000000',
+        value:
+          typeof walletBalance === 'number' ? (
+            formatMsatsToBtc(walletBalance)
+          ) : (
+            <Skeleton height='20px' />
+          ),
       },
     ],
-    [t]
+    [walletBalance, t]
   );
 
   return (
--- a/packages/utils/src/format.tsx
+++ b/packages/utils/src/format.tsx
@@ -1,3 +1,5 @@
+import { MSats } from './types';
+
 /**
  * Given a string, turn it into an "ellipsis sandwich" with the start and
  * end showing. If the text is shorter than it would be with an ellipsis,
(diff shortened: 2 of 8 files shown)


EthnTuttle:
> Opaque number types have been added for Btc, Sats, and MSats that makes it so that API types and functions can dictate which specific units they accept, to avoid the ambiguity of number not being clear what denomination it is

🧡

apps/guardian-ui/src/components/BalanceCard.tsx
```diff
@@ -1,21 +1,37 @@
...
   const { t } = useTranslation();
+  const { api } = useAdminContext();
+  const [auditSummary, setAuditSummary] = useState<AuditSummary>();
+
+  const walletBalance = auditSummary
+    ? auditSummary.module_summaries.wallet?.net_assets || (0 as MSats)
+    : undefined;
+
+  useEffect(() => {
+    api.audit().then(setAuditSummary).catch(console.error);
```
EthnTuttle, line 18: Just confirming I understand this, `useEffect` is used here because there's a network call being made?

packages/utils/src/format.tsx
```diff
@@ -9,3 +11,14 @@ export function formatEllipsized(text: string, size = 6) {
   }
   return `${text.substring(0, size)}...${text.substring(text.length - size)}`;
 }
+
+/**
+ * Given some number of msats, return a formatted string in the format of
+ * 0.0000000 or 1,234,5678.9000000
+ */
+export function formatMsatsToBtc(msats: MSats): string {
```
EthnTuttle, line 19: https://github.com/fedimint/ui/issues/182
Made this issue before I saw this. Changed Issue scope to writing tests for this so the use is more obvious.

packages/utils/src/types.ts
```diff
@@ -0,0 +1,5 @@
+// Opaque numeric types to avoid mixing different denominations of bitcoin
+type BitcoinUnit<K, T> = K & { _: T };
+export type Btc = BitcoinUnit<number, 'Btc'>;
+export type Sats = BitcoinUnit<number, 'Sats'>;
+export type MSats = BitcoinUnit<number, 'MSats'>;
```
EthnTuttle, line 5: Should we use bigint for sats and/or mSats?

apps/guardian-ui/src/components/BalanceCard.tsx
wbobeirne: `useEffect` is just a generic hook for saying "Run this every time any of my dependencies change." So in the case of this one, it'd run `api.audit()` any time `api` changes. Which never happens in practice here, but you can imagine if `audit` had an arugument like pagination, then it would run any time that updated too.

`useEffect` is often paired with network requests since you don't want them running every render, just every time some relevant piece of information changes, but that's not _all_ they're good for. They can also be useful for reacting to changes in component props, querying the DOM, or for binding to events on globals like `window`.

packages/utils/src/format.tsx
wbobeirne: Yep, would also be good to find all the old code where we're displaying sats or bitcoin and update it to use a set of common utility functions.

packages/utils/src/types.ts
wbobeirne: Sorry I should have written some notes about why I removed your `bigint` types and made them `number`. Using `bigint` isn't as simple as changing the type since it's an entirely separate class from `Number` and has runtime differences. JSON parsing of numbers will not automatically cast them as `BigInt`, they will initially be `Number`s which means we would have to do the class conversion ourselves, which means that we'd already be suffering from the accuracy loss at that point. The Rust codebase would have to serialize them as strings instead of numbers in order for us to preserve accuracy, and that would mean quite a few breaking changes.

I talked to Justin about this, but we both decided that it would probably be overkill. Here's why:
```ts
formatMsatsToBtc(Number.MAX_SAFE_INTEGER)
// '90,071.99254741'
```
You would need to have 90k bitcoin before you would start to see some numeric inaccuracies in your balances. While this could conceivably happen in a regtest environment, this is unlikely to ever happen even on testnet / mutinynet since it's hard to accrue that many coins. Even if that did happen, JavaScript continues to function with numbers well over the max "safe" integer, it just starts to introduce inaccuracies in the lower digits. So you'd start to see some rounding at the satoshi level. But we're not going to get any runtime crashes, and we _probably_ won't end up with crazy numbers.

Might be worth adding a note somewhere, I've made the mistake of going all strings / big numbers for bitcoin stuff under this impression in the past but it's just practically not worth the headache. But lmk if you have other concerns.

EthnTuttle:
Shoot. I believe this should have been merged into master.

wbobeirne:
That's OK, I'll open a new PR against master. Usually GitHub is smart enough to re-target the PR to master, not sure why it didn't happen here.

PR #177, opened 2023-09-01
Fix password logic

### LINKED ISSUE #134: Guardian UI: Cannot submit form after refresh if password set completes, submit fails
This is a regression, this definitely used to work. If you submit the form as a leader, and it sets your password but fails to submit configuration, your form will be bricked if you refresh. All subsequent attempts will attempt to run `api.setPassword`, which will fail with `Expected to be in AwaitingPassword state`.

We should properly be pulling the password out of session storage and initializing state with it.

wbobeirne:
Closes #134

### What This Does

* Don't save password to `sessionStorage` if `save_password` RPC failed
* Run `testPassword` during initial load check instead of assuming it's valid, set `needsAuth` if it fails
* Initialize `SetupContext`'s `password` state with `api.getPassword()`
  * Since this is set, we can simplify the logic for not re-submitting password during `submitConfiguration()`

### Steps to Test

1. Get to the setup configuration form as a Leader
2. Fill out the form with an invalid Bitcoin RPC URI and submit, confirm you get an error
3. Scroll up to the password input, confirm it has become read-only
4. Refresh the page, confirm the password input is filled in and read-only
6. Edit your `sessionStorage` in the developer tools and set the password to an invalid password and refresh, confirm you are asked to login
    * On `master`, you will be taken to the setup page with the invalid password unchecked
8. Enter the password and submit, confirm you're taken back to setup and the password field is filled in and read-only
9. Fill out the form with valid inputs, confirm the form submits and continues successfully
    * On `master`, the form will fail to submit with an error that says you're expected to be in `AwaitingPassword`

### DIFF
--- a/apps/guardian-ui/src/setup/SetupContext.tsx
+++ b/apps/guardian-ui/src/setup/SetupContext.tsx
@@ -130,8 +130,11 @@ export const SetupContextProvider: React.FC<SetupContextProviderProps> = ({
   initServerStatus,
   children,
 }: SetupContextProviderProps) => {
-  const [state, dispatch] = useReducer(reducer, initialState);
-  const { password, configGenParams, myName } = state;
+  const [state, dispatch] = useReducer(reducer, {
+    ...initialState,
+    password: api.getPassword() || initialState.password,
+  });
+  const { password, myName } = state;
   const [isPollingConsensus, setIsPollingConsensus] = useState(false);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


apps/guardian-ui/src/setup/SetupContext.tsx
```diff
@@ -237,9 +240,7 @@ export const SetupContextProvider: React.FC<SetupContextProviderProps> = ({
     useCallback(
       async ({ password: newPassword, myName, configs }) => {
         if (!password) {
-          if (!configGenParams) {
```
wbobeirne, line 240: This extra conditional was no longer necessary, now that we initialize `password` correctly.

EthnTuttle (approved):

PR #178, opened 2023-09-01
fix: gateway fonts

### LINKED ISSUE #139: Gateway UI missing fonts
Gateway uses the same ui package as guardian-ui, but the ui package relies on you having publicly accessible fonts at `/fonts/*.ttf`: https://github.com/fedimint/ui/blob/master/packages/ui/src/theme.tsx#L310-L330

We should probably make the paths to these fonts arguments for the `<Fonts />` component, so that each app can specify how to find its fonts. We should also probably move the fonts into an assets folder in `ui` so that all apps can reference the same file (See #126.)

wbobeirne:
Closes #139

### What This Does

* Moves font files into `packages/ui/assets` for sharing between gateway and guardian UIs
* Adds font src path props to `<Fonts />` instead of hard-coding a path
  * Both apps now do the `import` themselves to allow their respective bundlers to generate paths, then pass them as props to `<Fonts />`

### Steps to Test

1. Open the guardian UI, confirm both inter and space grotesk fonts load
1. Open the gateway UI, confirm both inter and space grotesk fonts load

## Screenshots

Screenshot 2023-09-01 at 3 56 10 PM


EthnTuttle (approved):
lgtm

PR #187, opened 2023-09-07
feat: add .envrc, offer users nix develop shell automatically

dpc:


wbobeirne (approved):
Ha, I didn't realize I had made this file and it was caught by the gitignore so it never made it in. I'd been using this the whole time!

PR #189, opened 2023-09-07
fix(guardian-ui): run consensus in VerifyGuardians before continuing

### LINKED ISSUE #169: Solo federation verification step skip puts UI in invalid state
Creating a solo federation looks like it works, but when you end up on the admin page at the end, none of the API requests work except status, and you'll see that you're in `VerifiedConfigs`. Refreshing will put you back at the verification step. We still need to call `startConsensus` before we can go to the admin UI.

https://github.com/fedimint/ui/assets/649992/2a35bac3-2ede-4ab8-a198-811c30a7e37c

wbobeirne:
Closes #169.

### What This Does

* Runs `api.runConsensus` before continuing from the `VerifyGuardians` step if you're a solo federation
  * Renders a spinner instead of the verify configs table while this is happening
  * The header and description copy is still kinda wrong since there are no other guardians, but we can refine this later

## Screenshots

https://github.com/fedimint/ui/assets/649992/fc40b642-e285-4dc7-8004-9c1cad30e7af

### DIFF
--- a/apps/guardian-ui/src/components/VerifyGuardians.tsx
+++ b/apps/guardian-ui/src/components/VerifyGuardians.tsx
@@ -13,6 +13,7 @@ import {
   Tag,
   useTheme,
   HStack,
+  CircularProgress,
 } from '@chakra-ui/react';
 import { CopyInput, FormGroup, Table } from '@fedimint/ui';
 import { useConsensusPolling, useSetupContext } from '../hooks';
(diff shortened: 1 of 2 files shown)


EthnTuttle (approved):

EthnTuttle:
@dpc can sleep soundly knowing this is fixed. 😅

PR #190, opened 2023-09-07
Guardian UI figma spec

EthnTuttle:


EthnTuttle:
@wbobeirne The previous PR went into this branch for the audit work.

wbobeirne:
Oops, missed this. Opened #191 instead, since it has cleaner history without the merge commit. Would you mind approving / merging that one?

PR #205, opened 2023-09-14
chore: sort and translate all i18n keys

### LINKED ISSUE #194: Sync & update all translations
Our translation files are all out of sync. All keys that have been added and removed from `en` should be reflected in their `es` and `ko` counterparts. We should investigate an automatic way of doing translations going forward to reduce the effort.

wbobeirne:
Closes #194

### What This Does

* Translates all missing keys for Korean and Spanish
  * Translation was done using https://cloud.google.com/translate
* Removes keys that appeared in translations that didn't appear in english
* Sorts translation keys alphabetically, except for `"common"` at the top

This was done manually using a script, I'll post a future PR with a version of this script that anyone should be able to use for future translations.


EthnTuttle (approved):
LGTM

Issue #157, opened 2023-08-28
DKG status error message renders `[object Object]`


wbobeirne:
Note that this issue is to fix the text being `[object Object]`, not to prevent any particular instance of this error happening.

<img width="490" alt="Screenshot 2023-08-28 at 12 55 47 PM" src="https://github.com/fedimint/ui/assets/649992/73c01cc3-2539-4e0d-bfc2-92319de70dd3">

PR #213, opened 2023-09-29
fix: DKG status error message renders

harsh9539:
Fixes: #157

### DIFF
--- a/apps/guardian-ui/src/components/RunDKG.tsx
+++ b/apps/guardian-ui/src/components/RunDKG.tsx
@@ -56,7 +56,7 @@ export const RunDKG: React.FC<Props> = ({ next }) => {
             setError(`${t('run-dkg.error-config')}`);
             break;
           default:
-            setError(`${t('run-dkg.error-default')} '${JSON.stringify(status)}'`);
+            setError(`${t('run-dkg.error-default')} '${JSON.stringify(status.server)}'`);
         }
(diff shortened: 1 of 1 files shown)


apps/guardian-ui/src/components/RunDKG.tsx
```diff
@@ -56,7 +56,7 @@ export const RunDKG: React.FC<Props> = ({ next }) => {
             setError(`${t('run-dkg.error-config')}`);
             break;
           default:
-            setError(`${t('run-dkg.error-default')} '${status}'`);
+            setError(`${t('run-dkg.error-default')} '${JSON.stringify(status)}'`);
```
wbobeirne, line 59: Thanks for the first time contribution @harsh9539! We don't actually want the whole object to be rendered here, it's quite large and contains many irrelevant fields. We just want the `status.server`.

```suggestion
            setError(`${t('run-dkg.error-default')} '${status.server}'`);
```

wbobeirne (changes requested):

apps/guardian-ui/src/components/RunDKG.tsx
harsh9539: I changed the file according to your need.

wbobeirne (approved):

harsh9539:
Thanku sir soo much

PR #218, opened 2023-10-06
gateway ui fixes

okjodom:
- removes quotations around deposit address
- shows gateway balance as sats (instead of msats)
- shows the correct balance available for withdrawal
- fixes number input in withdraw card
- enables 'withdraw all button' though this should not be used?

### DIFF
--- a/apps/gateway-ui/src/components/WithdrawCard.tsx
+++ b/apps/gateway-ui/src/components/WithdrawCard.tsx
@@ -4,6 +4,8 @@ import {
   Button,
   Input,
   InputGroup,
+  NumberInput,
+  NumberInputField,
   Modal,
   ModalBody,
   ModalCloseButton,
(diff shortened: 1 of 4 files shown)


apps/gateway-ui/src/components/WithdrawCard.tsx
```diff
@@ -115,29 +88,33 @@ export const WithdrawCard = React.memo(function WithdrawCard({
...
-              name='amount'
-            />
-
+            <NumberInput
+              min={0}
+              max={balanceMsat / 1000}
+              value={amount}
+              onChange={(value) => {
+                value && setAmount(parseInt(value));
+              }}
```
wbobeirne, line 99: I would avoid attempting to parse an int in state, since inputs work only on strings. We should probably keep this as a string and only parse when we're performing numeric operations. This can cause issues with sufficiently large numbers

Screenshot 2023-10-06 at 8 04 13 AM

wbobeirne: This is no worse than it currently is though, so it's fine

wbobeirne (approved):
Ahhh, much better.

apps/gateway-ui/src/components/WithdrawCard.tsx
okjodom: I see! 👌🏿

Issue #224, opened 2023-10-10
Replace environment variables with `config.json` loaded at runtime


wbobeirne:
@elsirion had a brilliant idea: Rather than build-time environment variables and find / replacing the template, we should just load a `/config.json` file. This will make it much easier to change the variables, and make it easier for other projects to server the UI (e.g. `fedimintd`)

The app will make a request to `/config.json` before executing any other code. This may require some refactoring to defer execution of various things until the fetch has happened, and to make configurable properties parameters for functions / classes where they're currently hard-coded.

EthnTuttle:
Does this mean runtime variable for the backend would be configured using `config.json`?
There is only one ENV VAR for the guardian-ui and two for the gateway-ui. Both have one that is the "connection string" for the API. The gateway then has a password, which should probably be refactored out of an ENV VAR.

EthnTuttle:
How do we want to handle changing of `config.json` during runtime? A hard refresh of the app might have it reconnect to a different mint than what was initialized with. These seems desirable but we'll have to account for it with the State management.

elsirion:
Re state mgmt: I noticed that properly restarting the setup process is hard/impossible. Even after deleting all state of `fedimintd`+restarting it the UI was still at the old state after a reload. I'd expect it to reinitialize its state from the federation after a reload.

EthnTuttle:
> Re state mgmt: I noticed that properly restarting the setup process is hard/impossible. Even after deleting all state of `fedimintd`+restarting it the UI was still at the old state after a reload. I'd expect it to reinitialize its state from the federation after a reload.

I haven't looked yet through the code but perhaps it can be asked during the dev call.
Is there a good API for this? Maybe a deep dive on setup API?

justinmoon:
> Re state mgmt: I noticed that properly restarting the setup process is hard/impossible. Even after deleting all state of `fedimintd`+restarting it the UI was still at the old state after a reload. I'd expect it to reinitialize its state from the federation after a reload.

Call: this might have been a side-effect of setting admin password via environment variable

EthnTuttle:
@wbobeirne 
Can you peak `config-json-reboot` branch? I feel I am close but would appreciate feedback if I'm going off course.

PR #229, opened 2023-10-11
Guardian UI: Terms of Service env var

### LINKED ISSUE #226: Configurable Terms of Service agreement
To cover the butts of Fedimint service providers, we'll want a way to configure a terms of service screen that guardians have to agree to before continuing setup. This will allow anyone who wants to host a setup to provide their own terms, but by default will not show up for people who run the code themselves.

wbobeirne:
Closes #226

### What This Does

* Adds a `REACT_APP_TOS` env var for guardian UI to allow fedimint hosts to add their own terms of service that guardians must agree to before setup
  * Adds a screen to the beginning of setup if this env var is present
  * Does nothing if env var is omitted or falsy
* Allows undefined env vars in `replace-react-env` script
  * Needed to allow `REACT_APP_TOS` to be optional, and we're going to remove this with #224 anyway

### Screenshots

Screenshot 2023-10-10 at 11 53 18 PM


justinmoon (approved):
Tested and it worked

PR #231, opened 2023-10-13
feat: Add updation of dockerhub description from README in workflow [Fixes #210]

### LINKED ISSUE #210: Update Docker hub description from README
https://github.com/peter-evans/dockerhub-description

EshaanAgg:
Fixes #210 

I have updated the `gateway-ui` and the `guardian-ui` images on DockerHub to update their descriptions from their respective READMEs. As there is no way to check GitHub actions locally, it is imperative that these workflow files are cross-verified before merging. 

Also, I'm thinking to add a `workflow-dispatch` option to these workflows. They would allow us to run these publishing to DockerHub actions as and when we want manually. Does this seem like functionality that we would like to have?

### DIFF
--- a/.github/workflows/gateway-ui.yml
+++ b/.github/workflows/gateway-ui.yml
@@ -48,6 +48,7 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
 
       - name: Update the Docker Hub description
+        if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
         uses: peter-evans/dockerhub-description@v3
         with:
           username: fedimintui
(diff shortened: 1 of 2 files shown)


EthnTuttle:
> Also, I'm thinking to add a `workflow-dispatch` option to these workflows. They would allow us to run these publishing to DockerHub actions as and when we want manually. Does this seem like functionality that we would like to have?

Please bring up in the Discord or during a Monday dev call.

EthnTuttle (approved):
LGTM. Will hold off on merging until a second set of eyes gets on it.

justinmoon:
Let's give it a try!

PR #233, opened 2023-10-14
feat: add config.json for runtime

EthnTuttle:
This removes an ENV VAR from the mprocs-nix-guardian.yml for testing. 
It makes turbo happy by adding "PUBLIC_URL" to dependencies.
It creates a default `config.json` file in the `public directory.
It loads aforementioned `config.json` in the constructor for the api.

If this pattern is acceptable, we can expand on what's available in config.json. But I imagine that will need a different pattern at that point. 🤷‍♂️


EthnTuttle:
#224

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -47,6 +47,37 @@ class BaseGuardianApi
...
+      this.websocketUrl = websocketUrlFromEnv;
+    } else {
+      this.retrieveWebsocketUrlFromConfig();
+    }
+  }
+
+  // Asynchronous method to fetch the websocketUrl from config.json
+  async retrieveWebsocketUrlFromConfig() {
+    const xhr = new XMLHttpRequest();
+    xhr.open('GET', process.env.PUBLIC_URL + '/config.json', false); // Synchronous request
```
wbobeirne, line 67: Never use synchronous XMLHttpRequest, it will block all execution while the request is in flight. This will result in a terrible user experience if we want to give the user any understanding of what's going on while this is loading.

Opening a socket is already asynchronous by nature, this doesn't need to be made synchronous to maintain the

apps/guardian-ui/public/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "fm_config_api": "ws://127.0.0.1:18138"
+}
```
wbobeirne, line 3: Rather than hard-coding this, let's have webpack provide it based on the env vars set. Otherwise this will be kind of difficult to modify after build.

EthnTuttle: Isn't the idea to be able to modify after build?

apps/guardian-ui/public/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "fm_config_api": "ws://127.0.0.1:18138"
```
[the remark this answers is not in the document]
EthnTuttle: So just upper case it?

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -47,6 +47,37 @@ class BaseGuardianApi
...
+    // If the environment variable is not set, try to fetch from config.json
+    if (websocketUrlFromEnv) {
+      this.websocketUrl = websocketUrlFromEnv;
+    } else {
+      this.retrieveWebsocketUrlFromConfig();
+    }
+  }
+
+  // Asynchronous method to fetch the websocketUrl from config.json
+  async retrieveWebsocketUrlFromConfig() {
```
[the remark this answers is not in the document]
EthnTuttle: It seemed appropriate to do in the AppContext but I didn't quite land on a good way to do that.

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -47,6 +47,37 @@ class BaseGuardianApi
...
+    const xhr = new XMLHttpRequest();
```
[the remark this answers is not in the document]
EthnTuttle: admittedly, I asked ChatGPT to write this part. 🫣

apps/guardian-ui/public/config.json
wbobeirne: I think we should hash out how this should work for all use-cases, since they're all quite different:

* Development
  * Typically env var based, not file editing based
  * Should have a default for people running manually, and for people using mprocs / devimint
  * Should _ideally_ also fetch `config.json` so that the development experience isn't drastically different from production
* Docker
  * Should allow for shelling in and modifying `config.json`
  * Should allow for extending the image and modifying the `config.json`
  * Should allow for providing env vars _at run time_ (not image build time)
* Static tarball
  * Should only allow for modifying `config.json`
  * Should **not** use environment variables from build time

With all these considered, I think we should never use `process.env.VAR` in typescript. Otherwise, builds will have those environment variables hard coded in. All code paths should fetch the `config.json`, the only difference should be how that file is populated.

I think for development, we'll either need to extend the webpack dev server (See https://create-react-app.dev/docs/proxying-api-requests-in-development/). For docker, we'll want to update the `config.json` when the container is run based on env vars. For the tarball, we should probably provide a default `config.json` but expect the user to update it with their own values.

PR #240, opened 2023-10-24
refactor: Ban HStack and VStack components

### LINKED ISSUE #198: Replace and ban `VStack` and `HStack` components
These component adds a `0.5rem` margin to every component after the first, making it difficult to understand when you combine it with flex's `gap` property. For instance, a `<VStack gap={4} />` will product an _actual_ gap of `6`, since `0.5rem` is equivalent to 8 pixels.

We should instead use `<Flex direction="row|column" />` since it doesn't add any default styling.

Bonus points for adding https://eslint.org/docs/latest/rules/no-restricted-imports for `HStack` and `VStack` after the refactor.

EshaanAgg:
Fixes #198 

The PR accomplishes the following:
- Remove all the instances of the `HStack` and `VStack` components from the codebase and replace the same with `Flex` with the same gap properties. 
- Add `ESLint` rules to ban these components, and add suggestion messages to use `Flex` instead.

### DIFF
--- a/apps/gateway-ui/src/App.tsx
+++ b/apps/gateway-ui/src/App.tsx
@@ -3,7 +3,6 @@ import {
   Box,
   Heading,
   Stack,
-  VStack,
   Text,
   Flex,
   useTheme,
@@ -93,10 +92,10 @@ export const App = React.memo(function Admin(): JSX.Element {
     <ApiProvider props={{ gateway }}>
       <Wrapper size='lg'>
         {error ? (
-          <VStack spacing={4}>
+          <Flex direction='column' gap={4}>
             <Heading size='md'>{t('common.error')}</Heading>
             <Text>{error}</Text>
-          </VStack>
+          </Flex>
         ) : (
           <Box>
             {gatewayInfo?.federations.length ? null : (
--- a/apps/gateway-ui/src/App.tsx
+++ b/apps/gateway-ui/src/App.tsx
@@ -92,7 +92,7 @@ export const App = React.memo(function Admin(): JSX.Element {
     <ApiProvider props={{ gateway }}>
       <Wrapper size='lg'>
         {error ? (
-          <Flex direction='column' gap={4}>
+          <Flex direction='column' gap={6}>
             <Heading size='md'>{t('common.error')}</Heading>
             <Text>{error}</Text>
           </Flex>
--- a/apps/gateway-ui/src/components/WithdrawCard.tsx
+++ b/apps/gateway-ui/src/components/WithdrawCard.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useState } from 'react';
 import {
   Box,
   Button,
+  Flex,
   Input,
   InputGroup,
   NumberInput,
@@ -15,7 +16,6 @@ import {
   ModalOverlay,
   Stack,
   Text,
-  VStack,
   useTheme,
 } from '@chakra-ui/react';
 import {
@@ -208,7 +208,11 @@ const ConfirmWithdrawModal = (
             <ModalHeader>{t('withdraw-card.confirm-withdraw')}</ModalHeader>
             <ModalCloseButton />
             <ModalBody>
-              <VStack alignItems='flex-start' justifyContent='space-between'>
+              <Flex
+                direction='column'
+                alignItems='flex-start'
+                justifyContent='space-between'
+              >
                 <Box>
                   <Text>{t('common.amount')}:</Text>
                   <Text>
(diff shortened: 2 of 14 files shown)


okjodom:
The lint rule banning these components doesn't seem to take effect in my dev testing: Validated by import and using `HStack` in both apps, then running lints

EshaanAgg:
@okjodom Updated the rule and all the occurrences as well. Please run `yarn clean && yarn build` before `yarn lint` to see the rule in action.

apps/gateway-ui/src/App.tsx
```diff
@@ -93,10 +92,10 @@ export const App = React.memo(function Admin(): JSX.Element {
     <ApiProvider props={{ gateway }}>
       <Wrapper size='lg'>
         {error ? (
-          <VStack spacing={4}>
+          <Flex direction='column' gap={4}>
```
wbobeirne, line 95: The gaps of things that were formerly `VStack` need to be increased, since `VStack` provided its own additional `margin`-based gaps.
```suggestion
          <Flex direction='column' gap={6}>
```

apps/gateway-ui/src/components/WithdrawCard.tsx
```diff
@@ -208,7 +208,11 @@ const ConfirmWithdrawModal = (
             <ModalHeader>{t('withdraw-card.confirm-withdraw')}</ModalHeader>
             <ModalCloseButton />
             <ModalBody>
-              <VStack alignItems='flex-start' justifyContent='space-between'>
+              <Flex
+                direction='column'
+                alignItems='flex-start'
+                justifyContent='space-between'
+              >
```
wbobeirne, line 215: This now needs a gap of `2`, since `VStack` was providing spacing
```suggestion
              <Flex
                direction='column'
                alignItems='flex-start'
                justifyContent='space-between'
                gap={2}
              >
```

wbobeirne (changes requested):
This is great, thank you! Only concern is that with this change we're messing with the spacing on all the existing components. They'll need to be adjusted to make sure that this PR only affects the code, and doesn't alter the actual padding of the apps.

Note that I just added comments to a couple examples, but _all_ of them will need their `gap`s adjusted.

apps/gateway-ui/src/App.tsx
EshaanAgg: Thanks got it. Will update the code accordingly!

apps/gateway-ui/src/components/WithdrawCard.tsx
EshaanAgg: Makes sense!

EshaanAgg:
@wbobeirne Updated the spacing as instructed by you. I believe that I caught it all the places, still you can let me know if I happened to miss any!

apps/guardian-ui/src/components/Login.tsx
```diff
@@ -40,8 +40,8 @@ export const Login: React.FC = () => {
 
   return (
     <form onSubmit={handleSubmit}>
-      <Flex direction='column' pt={8} gap={2} align='start' justify='start'>
-        <Flex direction='column' align='start' gap={2}>
+      <Flex direction='column' pt={8} gap={4} align='start' justify='start'>
+        <Flex direction='column' align='start' gap={4}>
```
wbobeirne, line 44: These were already `Flex` previously, so this has unnecessarily increased their `gap`

wbobeirne (changes requested):
Sorry, it looks like you increased too many of them. Many of these were `Flex` before, not `HStack` and `VStack`, so they've been increased when they should have had no changes. Can you only increase the ones that went from stack to flex?

apps/guardian-ui/src/components/Login.tsx
EshaanAgg: My bad. Fixing them right now.

wbobeirne: OMG so sorry! I thought I was looking at the full diff, but I guess I just opened the commit diff. So sorry! I'll review in full later.

EshaanAgg: No worries! Thanks for all the reviews and help!

wbobeirne (approved):
Found one little issue I pushed a fix for, any descendants of a `VStack` or `HStack` would have their margins overridden (part of the reason for this change!) But that caused a few of the elements that were children of them previously to have their previously overridden margins push stuff around some more. But I think I caught the main case of it.

Thanks a ton for this! Much easier to reason about spacing now.

PR #243, opened 2023-10-24
fix: configure gateway ui env from devimint env

okjodom:
When starting up gateway ui in devimint scenarios, we should configure REACT_APP_FM_GATEWAY_API and REACT_APP_FM_GATEWAY_PASSWORD to values set by devimint


EthnTuttle (approved):
Just needed yarn to be ENV VAR aware of what devimint is doing. Nice.

PR #245, opened 2023-10-25
fix: don't cache results of lint and test

okjodom:
Invalidates turbo cache on `@fedimint/eslint-config` and `@fedimint/tsconfig` dev dependency changes. This fixes a bug where `turbo lint` would require `turbo build` for the latest lint rules to take effect.

ref: https://turbo.build/repo/docs/core-concepts/caching

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,5 +1,10 @@
 {
   "$schema": "https://turbo.build/schema.json",
+  "globalDependencies": [
+    "packages/eslint-config/*.js",
+    "packages/eslint-config/*.json",
+    "packages/tsconfig/*.json"
+  ],
   "pipeline": {
     "build": {
       "outputs": ["build/**", "dist/**"],


wbobeirne (approved):
This is fine for now, but we could better optimize this by specifying which inputs to bust cache on. Otherwise we're just throwing out the turbo cache: https://turbo.build/repo/docs/core-concepts/caching/file-inputs

okjodom:
> This is fine for now, but we could better optimize this by specifying which inputs to bust cache on. Otherwise we're just throwing out the turbo cache: https://turbo.build/repo/docs/core-concepts/caching/file-inputs

nice. In this case, would we consider the whole `eslint-config` package as an input so any changes to the package busts lint cache?
📌

wbobeirne:
Simplest thing might be to make eslint config and tsconfig `globalDependencies` since every project uses them.

okjodom:
> Simplest thing might be to make eslint config and tsconfig `globalDependencies` since every project uses them.

works out nice :)

wbobeirne (approved):

PR #249, opened 2023-10-27
fix(guardian-ui): add getEnv utility for avoiding compiler env var optimizations

### LINKED ISSUE #248: Guardian UI docker image shows empty terms of service by default
I've seen some reports that running the 0.1.1+ docker image with no `REACT_APP_TOS` env var still shows the terms of service screen with an empty textarea, even if it wasn't set. It's possible there's some compile time optimization that is seeing that it's set to a non-empty value, and replacing the conditionals inline.

This would be fixed with #224 since it would all be runtime, but might be worth a fix before that lands since it's a more significant refactor.

wbobeirne:
Closes #248

### What This Does

Stringifies & parses env to avoid react-scripts compiler optimizations of environment variables. See comment in `guardian-ui/utils/env.ts` for more information on why this was done and how it works.

This bit of code is just a hack until #224 is implemented, since that wouldn't get any of the compiler optimizations since the config would be loaded dynamically.

### DIFF
--- a/apps/guardian-ui/src/GuardianApi.ts
+++ b/apps/guardian-ui/src/GuardianApi.ts
@@ -11,6 +11,7 @@ import {
   PeerHashMap,
   ModulesConfigResponse,
 } from './types';
+import { getEnv } from './utils/env';
 
 export interface SocketAndAuthInterface {
   // WebSocket methods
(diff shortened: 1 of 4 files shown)


justinmoon (approved):

justinmoon:
Should we backport this?

wbobeirne:
Yes, we'll want this and #250 to work for non-alephbft federations

PR #251, opened 2023-10-30
fix: gateway explorer links

### LINKED ISSUE #242: "view on mempool.space" and "view on amboss" don't work on mutinynet
"view on mempool.space" should point to mutinynet.com. Since I don't think there is a node explorer on mutinynet, I think we should just hide the "view on amboss" button.

image

okjodom:
When gateway is deployed on mainnet and testnet, gateway address and node links should point to valid mempool.space links.
On signet, we assume gateway is deployed on mutinynet.com where we point both links. When gateway is on regtest, or the network is not yet known, we don't show "view in explorer" links

Changes in gateway UI take effect after https://github.com/fedimint/fedimint/pull/3455 which introduces `network` field to gateway info. To validate these changes before the upstream changes modify the `network` prop in `FederationCard` and  observe how the links are affected. Valid network values are `"main" | "test" | "signet" | "regtest"`

closes #242

### DIFF
--- a/apps/gateway-ui/src/components/DepositCard.tsx
+++ b/apps/gateway-ui/src/components/DepositCard.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import {
   Text,
   Flex,
@@ -14,13 +14,16 @@ import { useTranslation } from '@fedimint/utils';
 import { GatewayCard } from './GatewayCard';
 import { ReactComponent as CopyIcon } from '../assets/svgs/copy.svg';
 import { ReactComponent as LinkIcon } from '../assets/svgs/linkIcon.svg';
+import { Network } from '../types';
 
 export interface DepositCardProps {
   federationId: string;
+  network?: Network;
 }
 
 export const DepositCard = React.memo(function DepositCard({
   federationId,
+  network,
 }: DepositCardProps): JSX.Element {
   const { t } = useTranslation();
   const { gateway } = React.useContext(ApiContext);
@@ -42,6 +45,11 @@ export const DepositCard = React.memo(function DepositCard({
           setError(message);
         });
   }, [address, federationId, gateway]);
+
+  const url = useMemo(
+    () => getAddressUrl(address, network),
+    [address, network]
+  );
 
   return (
     <GatewayCard
@@ -110,7 +118,7 @@ export const DepositCard = React.memo(function DepositCard({
           fontFamily={theme.fonts.body}
           _hover={{ textDecoration: 'underline' }}
           transition={`text-decoration 1s ease-in-out`}
-          href={`https://mempool.space/address/${address}`}
+          href={url.toString()}
           target='_blank'
           rel='noreferrer'
           w='fit-content'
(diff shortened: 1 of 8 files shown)


apps/gateway-ui/src/languages/es.json
EthnTuttle, line 1: Does this break or does it default to English?

okjodom: defaults to english since the translation key is renamed

EthnTuttle (dismissed):
LGTM.
Pending https://github.com/fedimint/fedimint/pull/3455, will merge.

okjodom:
> LGTM. Pending [fedimint/fedimint#3455](https://github.com/fedimint/fedimint/pull/3455), will merge.

Thought to clarify that this change hides the link when the `network` prop is undefined. IMO makes it okay to merge before upstream lands. When upstream lands and gateways update, the link will be shown per the network configured

apps/gateway-ui/src/languages/en.json
```diff
@@ -69,7 +69,7 @@
     "sort": "Sort"
   },
   "info-card": {
-    "amboss_node_link_text": "View on amboss.space",
+    "open-node-in-explorer": "Open in explorer",
```
wbobeirne, line 72: Not important, but we could change this to `View on {{origin}}` and update the `t` calls to include the origin of the link they pulled together. Not really worth holding up the PR, but ideally we avoid deviating from design wherever possible.

apps/gateway-ui/src/components/DepositCard.tsx
```diff
@@ -103,22 +106,37 @@ export const DepositCard = React.memo(function DepositCard({
...
-          href={`https://mempool.space/address/${address}`}
-          target='_blank'
-          rel='noreferrer'
-          w='fit-content'
-        >
-          {t('deposit-card.mempool_deposit_link_text')}
-        </Link>
-        <LinkIcon color={theme.colors.blue[600]} />
-      </Flex>
+      {network && network !== Network.Regtest && (
```
wbobeirne, line 109: Backwards compatibility could be maintained with always rendering this, and `default`ing to the mainnet link in the switch statement.

wbobeirne (commented):
As a point of policy, I'm also hesitant to merge changes that remove functionality with the current `master` of fedimint. I think when we want to target future changes, the game plan should be to maintain backwards compatibility (i.e. default to the mainnet link when `!network`) but priming the app to be ready for an incoming change when the API updates to get the new response type makes the most sense.

If you've got the time to make it backwards compatible, it'd be good to get into the habit of doing that. Otherwise we could just wait for the fedimint PR to land, there's no value in getting this into a `0.1.x` release unless a gateway release is also ported with this network property, which I suspect won't happen.

apps/gateway-ui/src/languages/en.json
okjodom: Interpolation with origin results in this madness `"View on https:&#x2F;&#x2F;mempool.space"`. Went with `View on {{host}}`

okjodom:
> If you've got the time to make it backwards compatible, it'd be good to get into the habit of doing that. Otherwise we could just wait for the fedimint PR to land, there's no value in getting this into a 0.1.x release unless a gateway release is also ported with this network property, which I suspect won't happen.

Updated the change to make it backward compatible in UI as a start

wbobeirne (approved):

Issue #254, opened 2023-11-01
Pre-fill verify hashes if status is `VerifiedConfigs`


wbobeirne:
If you enter all the verification hashes then refresh, it asks you to do it again but you're already in `VerifiedConfigs`. We should pre-fill the inputs for you if you're in that state.

justinmoon:
I think https://github.com/fedimint/ui/pull/256 closed this. Please re-open if I'm wrong.

PR #256, opened 2023-11-02
feat: prefills hashes if we're already in VerifiedConfigs

Kodylow:
addresses #254 , if user is in VerifiedConfigs prefills enteredHashes


okjodom (approved):
works great ✅

PR #257, opened 2023-11-02
feat: scroll to top on progress screen change

Kodylow:


okjodom:
We probably need to smoothen/animate the scroll to top. any feedback on this experience, @wbobeirne ?

justinmoon:
The current behavior here is very annoying. Also defer to @wbobeirne about how to fix.

Kodylow:
I can make it smooth I preferred just jumping to top, fine with either

wbobeirne:
I'll look at this one today, sorry for the delay.

wbobeirne:
Typical web behavior is that new pages jump to the top immediately on navigation, not smoothly, so that's what I'd advocate for. I think Kody's initial behavior was closer to a normal web experience.

wbobeirne (approved):

PR #268, opened 2023-11-07
guardian-ui: fix config verification

### LINKED ISSUE #255: Handle being in `VerifiedConfigs` when leader is in `ConsensusRunning`
We've got a potential bug where if you refresh the page after the leader has run `startConsensus`, you get stuck as a follower since you can no longer fetch their config. We're actually ready to go, we just need to call `startConsensus` at this point, but we have no UI to do it because it shows an error.

Not 100% sure the best thing to do here, automatically do it for the user? Show them a button?

okjodom:
At the config verification step after all the guardians have verified other configs, if the leader continued their own setup to completion before the follower guardians, there is a chance the followers would not be able to **resume and complete** their setup since we relied on their servers being able to continuously poll the leader's server for consensus configuration.

This PR fixes the bug, allowing all guardians to complete federation set-up in any order, provided they have successfully coordinated through config verification.

## Steps to validate

1. Before fix:
- Follow federation set-up through config verification until you see "All guardians have verified their codes"
- Reload any guardians setup page to confirm that the verified codes are automatically pre-filled on reload
- Complete set-up for the leader
- Reload the page for any other guardian and observe that config verification shows an error. The verification page expects the leader's server to still be in set up and not in consensus running

2. After fix:
- Follow federation set-up through config verification until you see "All guardians have verified their codes"
- Reload any guardians setup page to observe that the verified codes are automatically pre-filled on reload
- Complete set-up for the leader
- Reload the page for any other guardian and observe that the verified codes are automatically pre-filled on reload
- Confirm you can complete set-up for these "follower" guardians

Fixes #255

### DIFF
--- a/apps/guardian-ui/src/components/VerifyGuardians.tsx
+++ b/apps/guardian-ui/src/components/VerifyGuardians.tsx
@@ -37,7 +37,7 @@ export const VerifyGuardians: React.FC<Props> = ({ next }) => {
   const { t } = useTranslation();
   const {
     api,
-    state: { role, numPeers, peers },
+    state: { role, numPeers, peers, ourCurrentId },
   } = useSetupContext();
   const theme = useTheme();
   const isHost = role === GuardianRole.Host;
@@ -47,28 +47,27 @@ export const VerifyGuardians: React.FC<Props> = ({ next }) => {
   const [isStarting, setIsStarting] = useState(false);
   const [error, setError] = useState<string>();
 
-  // Poll for peers and configGenParams while on this page.
-  useConsensusPolling();
-
   const verifiedConfigs: boolean = peers.every(
     (peer) => peer.status === ServerStatus.VerifiedConfigs
   );
 
+  // Poll for peers and configGenParams while on this page.
+  useConsensusPolling(!verifiedConfigs);
+
   useEffect(() => {
     async function assembleHashInfo() {
-      try {
-        const [
-          {
-            consensus: { peers },
-            our_current_id,
-          },
-          hashes,
-        ] = await Promise.all([
-          api.getConsensusConfigGenParams(),
-          api.getVerifyConfigHash(),
-        ]);
+      if (peers.length === 0) {
+        return setError(t('verify-guardians.error'));
+      }
 
-        setMyHash(hashes[our_current_id]);
+      if (ourCurrentId === null) {
+        return setError(t('verify-guardians.error-peer-id'));
+      }
+
+      try {
+        const hashes = await api.getVerifyConfigHash();
+
+        setMyHash(hashes[ourCurrentId]);
         setPeersWithHash(
           Object.entries(peers)
             .map(([id, peer]) => ({
@@ -76,13 +75,13 @@ export const VerifyGuardians: React.FC<Props> = ({ next }) => {
               peer,
               hash: hashes[id as unknown as number],
             }))
-            .filter((peer) => peer.id !== our_current_id.toString())
+            .filter((peer) => peer.id !== ourCurrentId.toString())
         );
 
         // If we're already at the VerifiedConfigs state, prefill all the other hashes with the correct values
-        if (peers[our_current_id].status === ServerStatus.VerifiedConfigs) {
+        if (peers[ourCurrentId].status === ServerStatus.VerifiedConfigs) {
           const otherPeers = Object.entries(peers).filter(
-            ([id]) => id !== our_current_id.toString()
+            ([id]) => id !== ourCurrentId.toString()
           );
           setEnteredHashes(
             otherPeers.map(([id]) => hashes[id as unknown as number])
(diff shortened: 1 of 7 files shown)


apps/guardian-ui/src/utils/validators.ts
```diff
@@ -9,3 +11,9 @@ export const isValidNumber = (value: string, min?: number, max?: number) => {
 export const isValidMeta = (meta: [string, string][]) => {
   return meta.every(([key, value]) => key && value);
 };
+
+export function isConsesusParams(
+  params: ConfigGenParams | ConsensusParams
+): params is ConsensusParams {
+  return 'peers' in params;
+}
```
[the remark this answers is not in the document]
okjodom: good catch :)

apps/guardian-ui/src/components/VerifyGuardians.tsx
```diff
@@ -37,7 +37,7 @@ export const VerifyGuardians: React.FC<Props> = ({ next }) => {
   const { t } = useTranslation();
   const {
     api,
-    state: { role, numPeers, peers },
+    state: { role, numPeers, peers, ourCurrentId },
```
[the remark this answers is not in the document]
okjodom: nack / out of scope. `ourCurrentId` is a translation of server side `our_current_id` in consensus state related apis. I think we currently try to keep the UI and API types in sync by keeping name conventions

apps/guardian-ui/src/utils/validators.ts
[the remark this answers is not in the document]
okjodom: Afaik, defining consensus state as a union of `ConfigGenParams` and peer info makes it easier to understand and matches api state transitions, compared to having `ConfigGenParams` that may optionally have peer info.

Not strongly opinionated about it though

apps/guardian-ui/src/components/VerifyGuardians.tsx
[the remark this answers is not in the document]
Kodylow: good to go wasn't tracking

Kodylow (approved):

README.md
```diff
@@ -85,7 +85,7 @@ The gateway UI container is available at [`fedimintui/gateway-ui`](https://hub.d
 docker pull fedimintui/gateway-ui:0.1.0
 docker run \
   --platform linux/amd64 \
-  --env "REACT_APP_FM_GATEWAY_API='ws://127.0.0.1:8175'" \
+  --env "REACT_APP_FM_GATEWAY_API='http://127.0.0.1:8175'" \
   --env REACT_APP_FM_GATEWAY_PASSWORD=password \
```
Kodylow, line 89: can we set all these gateway passwords to "thereisnosecondbest" to keep in line with the fedimint docker images and setup guides we normally use?

wbobeirne: These are instructions for running a production release, so I would definitely want to avoid encouraging someone to continue to use that password on a real instance. It's meant to be replaced, not run as-is, so maybe we should make it something obvious like `[your-password-here]`

okjodom: +1 to @wbobeirne suggestion, and I'd follow up with these as doc updates given they're going to conflict with #274

okjodom: Dropped this tangential change to avoid conflicts with mentioned PR.

okjodom: #288

wbobeirne:
@okjodom I hit continue with the leader before verifying all configs on all followers, and got stuck on this screen:

CleanShot 2023-11-14 at 13 16 33@2x

Is this a regression from https://github.com/fedimint/fedimint/pull/3547? I was unable to get this to happen in `master`.

okjodom:
> @okjodom I hit continue with the leader before verifying all configs on all followers, and got stuck on this screen:
> 
> CleanShot 2023-11-14 at 13 16 33@2x
> 
> Is this a regression from [fedimint/fedimint#3547](https://github.com/fedimint/fedimint/pull/3547)? I was unable to get this to happen in `master`.

do you mean current repo `master`, or upstream `master`? trying a repro

wbobeirne:
Apologies, I meant `fedimint/ui@master` using the version of devimint pegged in the `flake.nix`, which was `23ee7cb6e96fce89bb024fbc1fcfccbfb3dc968b` at the time of my run.

okjodom:
> Apologies, I meant `fedimint/ui@master` using the version of devimint pegged in the `flake.nix`, which was `23ee7cb6e96fce89bb024fbc1fcfccbfb3dc968b` at the time of my run.

Figured I didn't properly update flake lock - now it's taking forever to recompile devimint after bump🤦🏿‍♂️

okjodom:
need to have this bug fixed

Kodylow:
Ran the tests with this yesterday closing leader's tab and refreshing everything, looks good to me.

Kodylow (approved):

okjodom:
thanks for the review. landing fix

Issue #217, opened 2023-10-05
Add docs for developing with nix


wbobeirne:
Our development docs are fairly docker-centric, but we should emphasize using nix / devimint more. We should have instructions that push people towards using `yarn run nix-guardian` and `yarn run nix-gateway`, but still have docker instructions as an alternative.

PR #274, opened 2023-11-08
Readme updates

lonestarr-btc:
1. Updated the Readme in the guardian-ui/apps directory to add the additional two guardian ui instances that are now spun up with docker compose.
2. In the repo root Readme, additional details were added on setting up Nix and running mprocs including a table to show the uri information for all instances spun up by the mprocs.yml file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Replace `-p 3001:3000` with a port of your choice, `REACT_APP_FM_GATEWAY_API` wi
 
 ## Development
 
-### Option 1 - Using Nix (preferred)
+### Option 1 - Running with Nix (preferred)
 
 1. Install Nix
    ```bash
@@ -131,7 +131,7 @@ nix-guardian spins up these instances
 | guardian-ui-3 | http://127.0.0.1:3002/ |
 | guardian-ui-4 | http://127.0.0.1:3003/ |
 
-### Option 2 - Using Docker Compose and Yarn manually
+### Option 2 - Running with Docker Compose and Yarn manually
 
 From root repo directory:
 
@@ -154,7 +154,7 @@ From root repo directory:
 
 After running through the config setup UI flow once, you will need to delete the `fedimintd` data to run through it again. To do this, delete the `fm_1`, `fm_2`, `fm_3`, and `fm_4` folder from the repo. These are data directories mounted to Docker containers running fedmintd and are listed in `.gitignore` so are safe to remove.
 
-### Running mprocs
+### Running with mprocs
 
 1. Install [mprocs](https://github.com/pvolok/mprocs)
 1. Run `mprocs -c mprocs.yml`
--- a/README.md
+++ b/README.md
@@ -36,6 +36,43 @@ Fedimint UI releases use semantic versioning (`major.minor.patch`)
 
 ## Running a Release
 
+### Running with Nix
+
+1. Install Nix
+   ```bash
+   curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
+   ```
+1. Add `experimental-features = nix-command flakes` to your `/etc/nix/nix.conf` file   
+1. In terminal `cd` to the fedimint-ui repo root directory
+1. Enter the following command to start Nix development environment
+   ```bash
+   nix develop
+   ```
+1. Run `yarn nix-gateway` or `nix-guardian` Note: **nix-gateway** preconfigures the federation so you don't have to go through federation setup. **nix-guardian** starts separate guardian nodes that are connected into the federation when you run through the federation setup process.
+
+**guardian-ui**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,9 +116,13 @@ From root repo directory:
...
 
-### Running with mprocs
+### Running with Nix and mprocs
 
 1. Install [mprocs](https://github.com/pvolok/mprocs)
+1. Install [Nix](https://nixos.org/download)
+1. Add `experimental-features = nix-command flakes` to your `/etc/nix/nix.conf` file
+1. In a terminal `cd` to the fedimint-ui repo root directory
+1. Run `nix develop`
 1. Run `mprocs -c mprocs.yml`
```
wbobeirne, line 126: This is actually a docker-centric process, the `nix` way of doing development is using the `yarn nix-guardian` and `yarn nix-gateway` commands. Those leverage `devimint` on your machine, they aren't using containers. The only value with of using nix with this set of instructions is that you wouldn't need to install `mprocs`.

The hope for #217 was for documenting usage of those `yarn nix-*` commands, which we have some light notes for in `### Running with local Fedimint`.

It might be wise to split the `## Development` section into two completely separate ones, one with nix and one with docker.

wbobeirne (commented):
Solid start (love the table) but hoping to really make the nix instructions very turn-key for people, since they don't need to do nearly as much as running with Docker. I'd also probably make sure it's in front of Docker instructions since it's such a nicer experience if you spend the effort to get it going.

README.md
lonestarr-btc: Got it. I'll keep at it. Thanks for the feedback. I'm still getting up to speed with Nix, but so too are many that will be reading this document, so hopefully that's a good thing.

README.md
```diff
@@ -116,9 +116,13 @@ From root repo directory:
 
 After running through the config setup UI flow once, you will need to delete the `fedimintd` data to run through it again. To do this, delete the `fm_1`, `fm_2`, `fm_3`, and `fm_4` folder from the repo. These are data directories mounted to Docker containers running fedmintd and are listed in `.gitignore` so are safe to remove.
 
-### Running with mprocs
+### Running with Nix and mprocs
 
 1. Install [mprocs](https://github.com/pvolok/mprocs)
+1. Install [Nix](https://nixos.org/download)
```
Kodylow, line 122: we also need to add the determinate systems downloader ( here as recommendation for macOS, normal nix download causes lots of problems whenever macOS updates 

`curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install`

README.md
Kodylow: yeah the steps are:

1. install nix
2. nix develop (just put up a pr that will also run yarn install when you do this so the build env is complete at this point)
3. yarn nix-gateway or yarn nix-guardian . Big difference is nix-gateway preconfigures the federation so you don't have to go through federation setup, nix-guardian starts the separate guardian nodes that you connect into a federation via the setup process. 

Will just added guardian UIs to `yarn nix-gateway` so you can see the guardian dashboards without doing setup, so might be worth renaming them to be more clear.

Kodylow (commented):
We should change the order of these things to put nix setup first and recommended. I know some frontend ppl don't like it and prefer docker but doing it with nix has gotten so much cleaner and easier now.

README.md
lonestarr-btc: Roger this. Will do.

README.md
lonestarr-btc: I think I got everything as you wanted. Appreciate the opportunity to help out. If you need me to dig into anything else I am here to serve.

lonestarr-btc: Kody, you beat me to the yarn install. You da man

README.md
```diff
@@ -36,7 +36,44 @@ Fedimint UI releases use semantic versioning (`major.minor.patch`)
...
+nix-guardian spins up these instances
+| instance | url |
+| ------------- | --- |
+| guardian-ui-1 | http://127.0.0.1:3000/ |
+| guardian-ui-2 | http://127.0.0.1:3001/ |
+| guardian-ui-3 | http://127.0.0.1:3002/ |
+| guardian-ui-4 | http://127.0.0.1:3003/ |
+
+
+### Option 2 - Build and Run from Source
```
okjodom, line 76: Can we make this the third option? IMO it's the most tedious to use

wbobeirne: Wait I think this is mistaken, the running with nix is meant for development, build and run from source and run with docker below are meant for running releases. The nix instructions should be under the `## Development` header on line 133, not under running a release.

lonestarr-btc: My bad, I moved them down into ## Development and reverted everything up in ## Running a release. Ready for review

okjodom (approved):

Issue #280, opened 2023-11-10
Low Block Confirmations Warning


Kodylow:
Guardian SetConfigurations should warn user when they try to set up a federation with <6 block confirmations.

PR #281, opened 2023-11-10
Block confs warning

Kodylow:
Addresses #280 

What should the recommended minimum be here and any recommendations on copy for the warning? Should I have it link to a more in-depth explanation in the fedimint docs?


okjodom (approved):
lgtm.  I think the warning is nice, and adequate?

wbobeirne (approved):
I think the copy is fine for now, we can expand with docs link once we have public docs for stuff like this.

PR #287, opened 2023-11-10
feat: gateway and guardian dashboard polling

### LINKED ISSUE #221: Add polling to guardian admin UI
We should be polling the socket for peer guardian statuses, epoch count, and balance.

Kodylow:
Fixes #221

### DIFF
--- a/apps/gateway-ui/src/App.tsx
+++ b/apps/gateway-ui/src/App.tsx
@@ -50,18 +50,24 @@ export const App = React.memo(function Admin(): JSX.Element {
           setLoading(false);
         });
     } else {
-      gateway
-        .fetchInfo()
-        .then((gatewayInfo) => {
-          setGatewayInfo(gatewayInfo);
-        })
-        .catch(({ message, error }) => {
-          console.error(error);
-          setError(message);
-        })
-        .finally(() => {
-          setLoading(false);
-        });
+      const fetchInfo = () => {
+        gateway
+          .fetchInfo()
+          .then((gatewayInfo) => {
+            setGatewayInfo(gatewayInfo);
+          })
+          .catch(({ message, error }) => {
+            console.error(error);
+            setError(message);
+          })
+          .finally(() => {
+            setLoading(false);
+          });
+      };
+
+      fetchInfo();
+      const interval = setInterval(fetchInfo, 5000);
+      return () => clearInterval(interval);
     }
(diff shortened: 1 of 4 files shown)


apps/gateway-ui/src/App.tsx
```diff
@@ -50,18 +50,24 @@ export const App = React.memo(function Admin(): JSX.Element {
...
+        gateway
+          .fetchInfo()
+          .then((gatewayInfo) => {
+            setGatewayInfo(gatewayInfo);
+          })
+          .catch(({ message, error }) => {
+            console.error(error);
+            setError(message);
+          })
+          .finally(() => {
```
okjodom, line 63: nit: We probably should remove `setLoading` state call from the polling fn

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -22,9 +22,15 @@ export const FederationAdmin: React.FC = () => {
 
   useEffect(() => {
     // TODO: poll server status
```
okjodom, line 24: nit: we can remove stale todo comment

apps/gateway-ui/src/App.tsx
Kodylow: fixed.

apps/guardian-ui/src/admin/FederationAdmin.tsx
Kodylow: fixed.

apps/gateway-ui/src/App.tsx
```diff
@@ -50,18 +49,21 @@ export const App = React.memo(function Admin(): JSX.Element {
...
-        .then((gatewayInfo) => {
-          setGatewayInfo(gatewayInfo);
-        })
-        .catch(({ message, error }) => {
-          console.error(error);
-          setError(message);
-        })
-        .finally(() => {
-          setLoading(false);
-        });
```
wbobeirne, line 64: I haven't tested but I'm pretty sure without this, `loading` will be stuck as true indefinitely.

okjodom: testing now

okjodom:
@Kodylow , I've pushed a small fix to the last commit

okjodom (approved):

PR #288, opened 2023-11-12
docs: explicitly encourage replacing app address and passwords

okjodom:
remove default application address and passwords from setup docs, to clarify need to fill-in  user specified values


Kodylow (approved):
LGTM

wbobeirne (approved):

PR #289, opened 2023-11-14
fix(guardian-ui): add missing qrcode.react package

wbobeirne:
Fixes a regression from #284, the `qrcode.react` module wasn't added to package.json causing docker image builds to fail.


Kodylow (approved):
LGTM sorry about that :/

PR #298, opened 2023-11-14
Role selector Mobile Updates

Kodylow:
Makes several changes to the Role Selector page:

Old:
image

New:

image

1. Updates to copy:
- Change to "Welcome, Guardian" to avoid stacking big "Fedimint" and make it easier to fit title on mobile screen
- Setup Leader: slarifies that it's just for setup, not a difference among guardians
- Follower: Clarifies their role as after setup leader and that that's when the actual federation gets set up.
- Removes mainnet warning: we should do the mainnet warnings on the leader configuration page if they actually select mainnet, not preemptively. Also tells ppl to file issues with us.

2. Moves warning below role selector, above next button:
- On mobile the big warning pushes the follower option below the fold, this looks and makes it more like an "acknowledge by clicking next"

3. Fixes mobile UI layout
- Fixes overflows in radiobuttongroup children
- centers header on mobile (will be centered across all mobile pages)
- 100% width for next button

### DIFF
--- a/apps/guardian-ui/src/languages/en.json
+++ b/apps/guardian-ui/src/languages/en.json
@@ -63,14 +63,14 @@
   },
   "role-selector": {
     "disclaimer-title": "Fedimint is alpha software",
-    "disclaimer-text": "You are using a very early version of Fedimint. It is recommended to not use mainnet bitcoin for your federation. Please use with caution.",
+    "disclaimer-text": "You are using a very early version of Fedimint. Please use with caution & report issues to the Fedimint developers on Github.",
     "leader": {
-      "label": "Leader",
-      "description": "Choose one of your Guardians as a Leader. The Leader will input information about the Federation."
+      "label": "Setup Leader",
+      "description": "Input configuration settings that other Guardians approve. One Guardian acts as Setup Leader. "
     },
     "follower": {
       "label": "Follower",
-      "description": "Guardian Followers (all other Guardians) will confirm information that the Leader inputs."
+      "description": "Once a Setup Leader sets the configurations, other Guardians choose this option to approve the settings and create the Federation."
     }
   },
   "run-dkg": {
@@ -117,7 +117,7 @@
         "title": "Terms of service"
       },
       "start": {
-        "title": "Welcome to Fedimint!",
+        "title": "Welcome, Guardian!",
         "subtitle": "Are you creating a Federation, or joining one?",
         "step": "Federation details"
       },
--- a/apps/guardian-ui/src/setup/FederationSetup.tsx
+++ b/apps/guardian-ui/src/setup/FederationSetup.tsx
@@ -118,12 +118,12 @@ export const FederationSetup: React.FC = () => {
           </Button>
         )}
         {title && (
-          <Heading size={['sm', 'md']} fontWeight={['medium']}>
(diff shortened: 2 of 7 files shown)


Kodylow:
I don't like the "Are you creating a federation or joining one" at all, it just seems unnecessary especially since the creating or joining aren't easily visually mapped to setup vs guardian from that position, I think we should drop it

packages/ui/src/RadioButtonGroup.tsx
```diff
@@ -71,28 +70,32 @@ export function RadioButtonGroup<T extends string | number>({
                 bg={theme.colors.blue[100]}
                 boxShadow={`0 0 0 6px ${theme.colors.blue[50]}`}
                 borderRadius='100%'
-                mixBlendMode='multiply'
+                // Ensure icons fit well and are not too large
+                p={1} // Adjust padding around the icon
               >
-                <Icon as={option.icon} />
+                <Icon as={option.icon} boxSize='24px' />
```
wbobeirne, line 76: This change makes them larger than the design, it has them at 20.

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -118,19 +118,17 @@ export const FederationSetup: React.FC = () => {
           </Button>
         )}
         {title && (
-          <Heading size='md' fontWeight='medium'>
+          <Heading size={['sm', 'md']} fontWeight={['medium']}>
```
wbobeirne, line 121: Nit, if it's not a responsive value you might as well just keep it a string

wbobeirne (commented):
Some small code notes in here.

I'm not sure how we want to come to consensus around copy changes. For the most part I agree with the idea behind these changes, but it's going to get hard to decide a source of truth for copy if we start calling audibles in PRs.

I'm pretty ok with these but want to give a sec for others to weigh in.

packages/ui/src/RadioButtonGroup.tsx
```diff
@@ -71,28 +70,32 @@ export function RadioButtonGroup<T extends string | number>({
...
               </Flex>
               <Flex
                 direction='column'
                 align='start'
                 flex={1}
                 minWidth={0}
-                wrap='wrap'
+                wrap='nowrap' // Changed to nowrap
```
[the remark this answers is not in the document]
Kodylow: fixed

apps/guardian-ui/src/languages/en.json
```diff
@@ -116,7 +116,7 @@
         "title": "Terms of service"
       },
       "start": {
-        "title": "Welcome to Fedimint!",
+        "title": "Welcome, Guardian!",
```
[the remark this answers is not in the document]
Kodylow: fixed and translated w google

apps/guardian-ui/src/setup/FederationSetup.tsx
Kodylow: fixed

packages/ui/src/RadioButtonGroup.tsx
Kodylow: fixed

wbobeirne (approved):

PR #299, opened 2023-11-14
fix: center header

Kodylow:
Centers header on all screens to match figma, also looks better on both desktop and mobile for gateway and guardian

image
image

image

### DIFF
--- a/packages/ui/src/Header.tsx
+++ b/packages/ui/src/Header.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
-import { Flex, Spacer } from '@chakra-ui/react';
+import { Flex } from '@chakra-ui/react';
 
 export const Header = React.memo(function Header() {
   return (
-    <Flex width='100%' mb='32px'>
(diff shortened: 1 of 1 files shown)


wbobeirne:
> also looks better on both desktop

idk man looking a little topsy turvy to in some spots to me, I think we should limit what we're changing to be limited in scope to guardian ui mobile setup unless we figure out a way to make it good everywhere

CleanShot 2023-11-14 at 15 13 43@2x

CleanShot 2023-11-14 at 15 15 55@2x

Kodylow:
@wbobeirne kk will limit centering to mobile and on the desktop setup flow

okjodom:
we need to pull the design folks back in :p

Kodylow:
I'm meeting with @skyler tomorrow we're going to fix all the figma stuff and put it in a single file for us to reference

wbobeirne (changes requested):
Going with #298 centering on mobile only instead

Issue #293, opened 2023-11-14
Update progression bar for Mobile UI


Kodylow:
Mobile UI should remove labels and radials, then center the header logo on each page.

Think we can also reduce the back button top margin?

![image](https://github.com/fedimint/ui/assets/74332828/98e0b85e-974d-43e3-bd2b-8edf706e3d83)

PR #300, opened 2023-11-14
feat: clean setup bar progress for mobile view

Kodylow:
Addresses #293 

Removes text for progress bar on mobile so it doesn't overlap. Will eventually change to mobile specific progress bar per figma designs without the radials:

old:

image

new:

image

### DIFF
--- a/apps/guardian-ui/src/components/SetupProgress.tsx
+++ b/apps/guardian-ui/src/components/SetupProgress.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Box, Center, Flex, Text, useTheme } from '@chakra-ui/react';
 import { ReactComponent as CheckIcon } from '../assets/svgs/white-check.svg';
 import { useTranslation } from '@fedimint/utils';
@@ -8,9 +8,10 @@ interface StepProps {
   text: string;
   state: StepState;
   isFirst?: boolean;
+  isMobile?: boolean;
 }
 
-const Step: React.FC<StepProps> = ({ text, state, isFirst }) => {
+const Step: React.FC<StepProps> = ({ text, state, isFirst, isMobile }) => {
   const theme = useTheme();
 
   const colorState = (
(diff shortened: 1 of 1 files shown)


apps/guardian-ui/src/components/SetupProgress.tsx
```diff
@@ -111,6 +114,16 @@ export const SetupProgress: React.FC<SetupProgressProps> = ({
...
+  const [isMobile, setIsMobile] = useState(window.innerWidth < 768);
+
+  useEffect(() => {
+    function handleResize() {
+      setIsMobile(window.innerWidth < 768);
+    }
+
+    window.addEventListener('resize', handleResize);
+    return () => window.removeEventListener('resize', handleResize);
+  }, []);
```
wbobeirne, line 126: Use https://chakra-ui.com/docs/hooks/use-media-query or https://chakra-ui.com/docs/hooks/use-breakpoint-value instead, they've already got this figured out.

apps/guardian-ui/src/components/SetupProgress.tsx
```diff
@@ -78,25 +79,27 @@ const Step: React.FC<StepProps> = ({ text, state, isFirst }) => {
...
+          color={colorState(
+            theme.colors.gray[700],
+            theme.colors.blue[600],
+            theme.colors.gray[700]
+          )}
+          whiteSpace='nowrap'
+        >
+          {text}
+        </Text>
+      )}
```
wbobeirne, line 102: Use https://chakra-ui.com/docs/components/show-hide instead.

apps/guardian-ui/src/components/SetupProgress.tsx
wbobeirne: But don't use either hook, see show / hide component below.

wbobeirne (changes requested):
This is a good quick win without having to do a full redesign. However you could do this much more simply using Chakra's own tools.

apps/guardian-ui/src/components/SetupProgress.tsx
Kodylow: fixed in following commit

Kodylow:
Chakra has a Progress bar component which does exactly what sahil was looking for, so just adjusted it using show/hide like you suggested to show on mobile:

image

apps/guardian-ui/src/components/SetupProgress.tsx
Kodylow: updated

wbobeirne (dismissed):
Dank, although Sahil's designs show rounded corners on the bar and use a different shade of blue. Is that something you could match up before merging? Otherwise LGTM.

Kodylow:
Updated for darker progress bar (theme.blue[1000]) and rounded borders:

image

Although now it's a different color than the desktop one which is theme.blue[600] , will address with skyler in design meeting tomorrow to make sure figma reflects reality

wbobeirne (approved):

PR #302, opened 2023-11-15
Set configuration leader options mobile updates and form headers

Kodylow:
Draft: can't get the isvalid states working properly for the different form groups.


Kodylow:
Closing in favor of #303

PR #303, opened 2023-11-15
Leader  UI mobile updates

Kodylow:
Draft: can't get the isvalid states working properly for the different form groups.


Kodylow:
closing in favor of #310

Issue #295, opened 2023-11-14
Remove "Sharing info between Guardians" Copy - just "Success"


Kodylow:
This copy specifically takes up a ton of space on mobile, and it's redundant given the "Running some numbers" copy below:

![image](https://github.com/fedimint/ui/assets/74332828/4a3b813e-b1e5-43cf-8570-95ca34bd763f)

wbobeirne:
> just "Boom"

I think this is gonna be pretty confusing to non-English speakers (and honestly, kinda confusing to English speakers as well.) Boom kinda sounds like something bad happened 😅

Kodylow:
Can we change to "Success" then?

PR #305, opened 2023-11-15
fix: update success copy

Kodylow:
Addresses #295  (sorry for shitty screenshot it's only there for a half second)

image


apps/guardian-ui/src/languages/es.json
```diff
@@ -128,7 +128,7 @@
       },
       "run-dkg": {
         "subtitle": "Todos los Guardianes han validado los detalles de la configuración de la Federación. Realizando algunos cálculos...",
-        "title": "¡Boom! Compartiendo información entre los Guardianes"
+        "title": "¡Éxito!"
```
okjodom, line 131: i'd remove these entries and let the bot auto translate for us so we have a consistent source of translations

okjodom: for es and ko

wbobeirne: We don't yet have the auto-translation code checked in so we don't yet have a canonical auto-translator. I think until then doing it ad-hoc is fine, but we should really get that formalized.

okjodom: #312

okjodom (approved):
okay on copy change

PR #311, opened 2023-11-16
fix: lightning gateway dashboard changes

Kodylow:
Made some copy changes per discussions from design call earlier:

1. Change top title to "Lightning Gateway Dashboard" to make it more clear (currently easily confused with guardian dashboard)
2. Move federation name onto balance card
3. Add clarifications to deposit and withdraw for which direction ecash is going, people have found that confusing "So if I deposit bitcoin then that means I get ecash?", it's unintuitive.

image

### DIFF
--- a/apps/gateway-ui/src/components/BalanceCard.tsx
+++ b/apps/gateway-ui/src/components/BalanceCard.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import { Text, useTheme } from '@chakra-ui/react';
 import { useTranslation } from '@fedimint/utils';
-import { GatewayCard } from './GatewayCard';
+import { FedNameCard } from './FedNameCard';
 
 interface BalanceCardProps {
   balance_msat: number;
+  federationName: string;
 }
 
 export const BalanceCard = React.memo(function BalanceCard(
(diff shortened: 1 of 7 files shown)


apps/gateway-ui/src/languages/en.json
```diff
@@ -66,17 +67,18 @@
...
-    "card_header": "Node ID"
+    "card_header": "Lightning Node ID"
   },
   "withdraw-card": {
     "address-label": "Your address:",
     "address-placeholder": "Enter your btc address",
     "amount-label": "Amount (sats):",
     "amount-placeholder": "Enter amount in sats",
-    "card-header": "Withdraw Bitcoin",
+    "card-header": "Withdraw Bitcoin (eCash -> BTC)",
```
okjodom, line 81: don't think we should translate `(eCash -> BTC)

apps/gateway-ui/src/components/BalanceCard.tsx
```diff
@@ -15,18 +16,21 @@ export const BalanceCard = React.memo(function BalanceCard(
...
+      </Text>
       <Text
-        fontSize='24px'
+        fontSize='xl'
         fontWeight='600'
         color={theme.colors.gray[800]}
         fontFamily={theme.fonts.body}
       >
-        {balance_msat / 1000}
+        {balance_msat / 1000} {'sats'}
```
okjodom, line 32: nit: `${balance_msat / 1000} sats`

okjodom (approved):
lgtm

PR #315, opened 2023-11-20
refactor: simplify guardian api implementation

### LINKED ISSUE #176: Simplify `GuardianApi` types & implementation
From Discord:

> Looking at the `GuardianApi` implementation, is there a reason we split this out into a `BaseGuardianApi` and `GuardianApi`  class, and why we have all these separate interfaces that the classes implement (`implements SocketAndAuthInterface, RpcInterface, SharedApiInterface`)? It feels a little unnecessarily abstracted for what's ultimately just going to be a single API implementation, and requires a lot of duplication of methods.

We can just have a single `GuardianApi` class, and derive its types from the class implementation, rather than defining an interface and having it implement that. That way adding, removing, or updating an RPC method only requires changes in one place.

okjodom:
makes guardian api type declaration and implementation simpler 🖌️ 

closes #176

### DIFF
--- a/apps/guardian-ui/src/types.tsx
+++ b/apps/guardian-ui/src/types.tsx
@@ -124,3 +124,39 @@ export type SetupAction =
       type: SETUP_ACTION_TYPE.SET_OUR_CURRENT_ID;
       payload: number;
     };
+
+// Setup RPC methods (only exist during setup)
+export enum SetupRpc {
+  setPassword = 'set_password',
+  setConfigGenConnections = 'set_config_gen_connections',
+  getDefaultConfigGenParams = 'default_config_gen_params',
+  getConsensusConfigGenParams = 'consensus_config_gen_params',
+  setConfigGenParams = 'set_config_gen_params',
+  runDkg = 'run_dkg',
+  verifiedConfigs = 'verified_configs',
+  startConsensus = 'start_consensus',
+  restartSetup = 'restart_federation_setup',
+}
+
+// Admin RPC methods (only exist after run_consensus)
+export enum AdminRpc {
+  version = 'version',
+  fetchBlockCount = 'block_count',
+  federationStatus = 'status',
+  inviteCode = 'invite_code',
+  config = 'client_config',
+  modulesConfig = 'modules_config_json',
+  moduleApiCall = 'module',
+  audit = 'audit',
+}
+
+export enum SharedRpc {
+  auth = 'auth',
+  status = 'status',
+  getVerifyConfigHash = 'verify_config_hash',
+}
+
+export enum ModuleRpc {
+  listGateways = 'list_gateways',
+  blockCount = 'block_count',
+}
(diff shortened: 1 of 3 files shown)


okjodom:
@wbobeirne pls take a look when you have the time. hopefully we can build on top of this for https://github.com/fedimint/ui/issues/316

apps/guardian-ui/src/types.tsx
```diff
@@ -123,3 +134,37 @@ export type SetupAction =
...
+
+export interface AdminApiInterface extends SharedApiInterface {
+  version: () => Promise<Versions>;
+  fetchBlockCount: () => Promise<number>;
+  inviteCode: () => Promise<string>;
+  config: (connection: string) => Promise<ConfigResponse>;
+  audit: () => Promise<AuditSummary>;
+  modulesConfig: () => Promise<ModulesConfigResponse>;
+  moduleApiCall: <T>(moduleId: number, rpc: LightningModuleRpc) => Promise<T>;
+}
```
wbobeirne, line 166: One of the things I was hoping to get rid of here was the double typing of every method. How would you feel about replacing this with `Pick`ing the methods off the `GuardianApi`?
```suggestion
type SharedApiMethods = 'auth' | 'status' | 'getVerifyConfigHash';
type SetupApiMethods =
  | 'setPassword'
  | 'setConfigGenConnections'
  | 'getDefaultConfigGenParams'
  | 'getConsensusConfigGenParams'
  | 'setConfigGenParams'
  | 'runDkg'
  | 'verifiedConfigs'
  | 'startConsensus';
type AdminApiMethods =
  | 'version'
  | 'fetchBlockCount'
  | 'inviteCode'
  | 'config'
  | 'audit'
  | 'modulesConfig'
  | 'moduleApiCall';

type SharedApiInterface = Pick<GuardianApi, SharedApiMethods>;

export type SetupApiInterface = SharedApiInterface &
  Pick<GuardianApi, SetupApiMethods>;
export type AdminApiInterface = SharedApiInterface &
  Pick<GuardianApi, AdminApiMethods>;
```

You could then remove `implements SetupApiInterface, AdminApiInterface` from the `class GuardianApi` and now it's the single source of truth for method implementation.

wbobeirne: Oh wait even better, move the `SharedRpc`, `SetupRpc`, and `AdminRpc` enums into types and just `Pick` from those. Super single source of truth!

wbobeirne (approved):
This is already an improvement so I'm ok with it going in as-is, but some suggestions that could make this even better.

apps/guardian-ui/src/types.tsx
wbobeirne: Nah, you can do `keyof typeof Enum` instead of `Enum` to get the camel cased versions that we use.

okjodom: ack. did a refresh of PR and applied type defs using enums

Kodylow (approved):

PR #319, opened 2023-11-24
fix: simplify footer w/skyler changes

Kodylow:
image
image


okjodom (approved):
🎆 for quick clickthrough, it'd be cool to link the updated figma in description

Issue #320, opened 2023-11-27
Top Level Indicators for Bitcoin Options


Kodylow:
Guardian setup UI should show which bitcoin network is currently selected for the configuration options, it defaults to mutinynet and closed so a user wouldn't know what network they're on.

Kodylow:
addressed by #321

PR #321, opened 2023-11-27
feat: add network text on bitcoin settings top bar

Kodylow:
image

image

image

addresses #320


apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -277,7 +277,16 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
             />
           </FormGroup>
         )}
-        <FormGroup icon={BitcoinLogo} title='Bitcoin settings' isOpen={false}>
+        <FormGroup
+          icon={BitcoinLogo}
+          title={
+            <>
+              <span>Bitcoin settings: </span>
```
okjodom, line 284: needs translation

apps/guardian-ui/src/utils/network.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+export const getNetworkIndicator = (network: string, bitcoinRpcUrl: string) => {
```
okjodom, line 3: wondering if we have need for a similar util/ux in gateway UI?

okjodom (dismissed):
ack. ✅

apps/guardian-ui/src/utils/network.tsx
Kodylow: no that one already shows the network, can add in a separate PR if we want it there too

Kodylow: added it to the guardian dashboard at least

Kodylow:
image

packages/ui/src/network.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+export const getNetworkIndicator = (network: string, bitcoinRpcUrl: string) => {
```
wbobeirne, line 3: Any reason this was made a function and not a component?

packages/ui/src/network.tsx
```diff
@@ -0,0 +1,35 @@
...
+      if (bitcoinRpcUrl === 'https://mutinynet.com/api') {
+        color = 'red';
+        name = 'Mutinynet';
+      } else {
+        color = 'purple';
+        name = 'Signet';
+      }
+      break;
+    case 'regtest':
+      color = 'skyblue';
```
wbobeirne, line 26: This has really rough readability on white, can we pick a darker color?

packages/ui/src/NetworkIndicator.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  switch (network) {
+    case 'bitcoin':
+      color = '#FF9900';
+      name = 'Mainnet';
+      break;
+    case 'testnet':
+      color = '#6BED33';
+      name = 'Testnet';
+      break;
```
wbobeirne, line 23: Seems like these are coming back "Unknown" instead

CleanShot 2023-11-30 at 09 57 40@2x
CleanShot 2023-11-30 at 09 57 35@2x

wbobeirne (changes requested):

PR #322, opened 2023-11-28
WIP: update to 0.2 fedimint dependencies

justinmoon:
Me and @Kodylow did some work to get the UI to support fedimint [0.2 release branch](https://github.com/fedimint/fedimint/tree/releases/v0.2).

To make testing faster we cherry-picked some WIP code to make single guardians setups work. These changes should be extracted into a separate PR.

### DIFF
--- a/apps/gateway-ui/src/GatewayApi.ts
+++ b/apps/gateway-ui/src/GatewayApi.ts
@@ -117,13 +117,13 @@ export class GatewayApi {
 
   requestWithdrawal = async (
     federationId: string,
-    amountSat: number,
+    msatAmountOrAll: number | string, // you can pass msat amount or 'all'
     address: string
   ): Promise<string> => {
     try {
       const res: Response = await this.post('withdraw', {
         federation_id: federationId,
-        amount: amountSat,
+        amount: msatAmountOrAll,
         address,
       });
 
--- a/apps/gateway-ui/src/components/WithdrawCard.tsx
+++ b/apps/gateway-ui/src/components/WithdrawCard.tsx
@@ -19,6 +19,7 @@ import {
   useTheme,
   Link,
   Icon,
+  Checkbox,
 } from '@chakra-ui/react';
 import { MSats, Network } from '@fedimint/types';
 import {
@@ -47,6 +48,7 @@ export const WithdrawCard = React.memo(function WithdrawCard({
   const [modalState, setModalState] = useState<boolean>(false);
   const [address, setAddress] = useState<string>('');
   const [amount, setAmount] = useState<number>(0);
+  const [withdrawAll, setWithdrawAll] = useState<boolean>(false);
   const [withdrawTxUrl, setWithdrawTxUrl] = useState<URL | null>(null);
 
   const handleChangeAddress = (ev: React.ChangeEvent<HTMLInputElement>) => {
@@ -64,7 +66,7 @@ export const WithdrawCard = React.memo(function WithdrawCard({
   };
 
   const createWithdrawal = useCallback(() => {
-    if (amount <= 0) {
+    if (!withdrawAll && amount <= 0) {
       return setError(`${t('withdraw-card.error-amount')}`);
     }
     if (!address) {
@@ -72,11 +74,12 @@ export const WithdrawCard = React.memo(function WithdrawCard({
     }
     setError('');
     setModalState(true);
-  }, [amount, address, t]);
+  }, [withdrawAll, amount, address, t]);
 
   const startWithdrawal = useCallback(() => {
+    const withdrawalAmount = withdrawAll ? 'all' : amount;
     gateway
-      .requestWithdrawal(federationId, amount, address)
+      .requestWithdrawal(federationId, withdrawalAmount, address)
       .then((txId) => {
         setWithdrawTxUrl(getTxUrl(txId));
         setAddress('');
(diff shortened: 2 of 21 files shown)


Kodylow:
cleaned up in #323 to remove the solo guardian stuff

justinmoon:
> cleaned up in #323 to remove the solo guardian stuff

Need to port some of my newer changes to your PR ...

apps/gateway-ui/src/GatewayApi.ts
```diff
@@ -117,13 +117,13 @@ export class GatewayApi {
 
   requestWithdrawal = async (
     federationId: string,
-    amountSat: number,
+    msatAmountOrAll: number | string, // you can pass msat amount or 'all'
```
justinmoon, line 120: This currently fails because fedimint expects `All`, not `all`. Trying to fix this deserialization in https://github.com/fedimint/fedimint/pull/3755 so we can have more consistent snake_case API.

apps/gateway-ui/src/components/WithdrawCard.tsx
```diff
@@ -111,9 +114,11 @@ export const WithdrawCard = React.memo(function WithdrawCard({
             <NumberInput
               min={0}
               max={balanceMsat / 1000}
-              value={amount}
+              // FIXME: we should subtract expected fees here in withdrawAll case
+              value={withdrawAll ? balanceMsat / 1000 : amount}
```
justinmoon, line 118: This shows your whole balance, and should instead show `balance - expected_fees`

apps/gateway-ui/src/languages/es.json
```diff
@@ -82,7 +82,7 @@
     "total_bitcoin": "Bitcoins totales:",
     "view": "Vista",
     "withdraw": "Retirar",
-    "withdraw_all": "Retirar todo",
+    "withdraw-all": "Retirar todo",
```
justinmoon, line 85: trying to have consistent keys with dashes

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -181,10 +182,10 @@ class BaseGuardianApi
 enum SetupRpc {
   setPassword = 'set_password',
   setConfigGenConnections = 'set_config_gen_connections',
-  getDefaultConfigGenParams = 'get_default_config_gen_params',
-  getConsensusConfigGenParams = 'get_consensus_config_gen_params',
+  getDefaultConfigGenParams = 'default_config_gen_params',
```
justinmoon, line 185: We went through and checked names of all RPCs

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -366,8 +371,20 @@ export class GuardianApi
...
+        )?.[0]
+      : undefined;
+
+    if (!walletModuleId) {
+      throw new Error('No wallet module found');
+    }
+    return this.moduleApiCall(
+      Number(walletModuleId),
+      WalletModuleRpc.blockCount
+    );
```
justinmoon, line 387: For module api calls I think it's better to pass the config in as an argument like this in order to find the instance ID. https://github.com/fedimint/ui/blob/3291e5b01e10ac46bd65497f79068079cb3f409c/apps/guardian-ui/src/components/GatewaysCard.tsx#L35-L50 has this sort of logic in the view layer which doesn't seem right.

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -34,15 +34,18 @@ export const FederationAdmin: React.FC = () => {
 
   useEffect(() => {
     if (!inviteCode) return;
-    api.config(inviteCode).then(setConfig).catch(console.error);
+    api.config().then(setConfig).catch(console.error);
```
justinmoon, line 37: It's a little weird that we can just fetch the config without an invite code cc @joschisan @elsirion. I understand we can update this behavior in the future but doesn't seem quite right to me.

apps/guardian-ui/src/components/BalanceCard.tsx
```diff
@@ -10,9 +10,12 @@ export const BalanceCard: React.FC = () => {
...
   const [auditSummary, setAuditSummary] = useState<AuditSummary>();
 
+  // FIXME: we shouldn't default to 0 balance
   const walletBalance = auditSummary
-    ? auditSummary.module_summaries.wallet?.net_assets || (0 as MSats)
-    : undefined;
+    ? Object.entries(auditSummary.module_summaries).find(
+        (m) => m[1].kind === ModuleKind.Wallet
+      )?.[1].net_assets
+    : (0 as MSats);
```
justinmoon, line 18: We really shouldn't show incorrect balances, ever. If we don't get the data we expect, I think we should display an error instead.

apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -378,16 +395,17 @@ export class GuardianApi
     return this.base.call(AdminRpc.inviteCode);
   };
 
-  config = (connection: string): Promise<ConfigResponse> => {
-    return this.base.call<ConfigResponse>(AdminRpc.config, connection);
+  // TODO: FIXME
```
justinmoon, line 398: this fixme can be removed

okjodom:
This draft is replaced by #323 right?

justinmoon:
> This draft is replaced by #323 right?

Yes. This also has some WIP code to fix "withdraw all" in the gateway UI. I'll close this and extract remaining code into a new PR.

Issue #316, opened 2023-11-22
Revise guardian api endpoints to match upstream changes


okjodom:
Guardian api endpoints have changed upstream and we need to update these in UI repo. A good cutoff point for this might be https://github.com/fedimint/fedimint/pull/3658 , so we bump fedimint to a version that includes all changes to [this file](https://github.com/fedimint/fedimint/blob/master/fedimint-core/src/endpoint_constants.rs)

PR #323, opened 2023-11-28
feat: support fedimint-0.2

Kodylow:
Closes https://github.com/fedimint/ui/issues/316

### DIFF
--- a/apps/guardian-ui/src/GuardianApi.ts
+++ b/apps/guardian-ui/src/GuardianApi.ts
@@ -1,7 +1,7 @@
 import { JsonRpcError, JsonRpcWebsocket } from 'jsonrpc-client-websocket';
 import {
   AuditSummary,
-  ConfigResponse,
+  ClientConfig,
   ConfigGenParams,
   ConsensusState,
   FederationStatus,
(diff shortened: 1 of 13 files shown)


apps/guardian-ui/src/GuardianApi.ts
```diff
@@ -378,16 +392,17 @@ export class GuardianApi
     return this.base.call(AdminRpc.inviteCode);
   };
 
-  config = (connection: string): Promise<ConfigResponse> => {
-    return this.base.call<ConfigResponse>(AdminRpc.config, connection);
+  // TODO: FIXME
```
wbobeirne, line 395: TODO: remove me

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -34,15 +34,18 @@ export const FederationAdmin: React.FC = () => {
 
   useEffect(() => {
     if (!inviteCode) return;
-    api.config(inviteCode).then(setConfig).catch(console.error);
+    api.config().then(setConfig).catch(console.error);
   }, [inviteCode, api]);
 
+  // TODO: FIXME to use better loading state
+  if (!config || !status) return <Text>{'Loading...'}</Text>;
```
wbobeirne, line 41: If `FederationInfoCard` allows for `config` to be `undefined`, we don't need this early return at all. Probably better to delegate each component to handle its own loading state instead of one huge one.

apps/guardian-ui/src/admin/FederationAdmin.tsx
```diff
@@ -34,15 +34,18 @@ export const FederationAdmin: React.FC = () => {
 
   useEffect(() => {
     if (!inviteCode) return;
-    api.config(inviteCode).then(setConfig).catch(console.error);
+    api.config().then(setConfig).catch(console.error);
   }, [inviteCode, api]);
```
wbobeirne, line 38: Since we don't need `inviteCode`, can we just call config right away without waiting to fetch it?

wbobeirne (approved):

wbobeirne (approved):

justinmoon:
Oh nice you updated docker images as well 👍. Just occurred to me that we probably forgot that!

wbobeirne:
Yep, might want to add a script for updating the hash since it's in so many places.

PR #329, opened 2023-11-30
fix: mutinynet url matching

Kodylow:
Noticed that if you have a trailing slash on the https://mutinynet.com/api/ rpc it doesn't register as mutinynet, so this converts to URL to match against trailing slash and http


packages/ui/src/NetworkIndicator.tsx
```diff
@@ -12,8 +12,6 @@ export const NetworkIndicator: FC<NetworkIndicatorProps> = ({
   let color: string;
   let name: string;
 
-  console.log({ network, bitcoinRpcUrl });
```
wbobeirne, line 15: Ughh sorry this is me, we should add a no console log linter

wbobeirne (changes requested):
`new URL` will throw if it's not a valid URL. Do the URL constructor inside of the component, not at the call site, and wrap in a try / catch.

Kodylow:
updated and refactored out to separate the check.

packages/ui/src/NetworkIndicator.tsx
```diff
@@ -5,43 +5,39 @@ interface NetworkIndicatorProps {
...
+    default: { color: 'gray', name: 'Unknown' },
+  };
+
+  return networkDetails[network] || networkDetails['default'];
+};
+
+const isMutinynet = (bitcoinRpcUrl: string) => {
+  try {
+    const url = new URL(bitcoinRpcUrl);
+    return url.host === 'mutinynet.com' && url.pathname === '/api';
```
wbobeirne, line 27: This still doesn't work, do you just want to drop the pathname and make it mutinynet if it's from mutinynet.com?

CleanShot 2023-12-05 at 12 19 13@2x

wbobeirne (changes requested):

wbobeirne (approved):

PR #331, opened 2023-12-01
fix: blockcount bug + yarn format

Kodylow:
image

Block Count is confusing, Should be "Consensus Block Height"


justinmoon (approved):

justinmoon:
Maybe we should just transform count to height since that's what shows up on block explorers and stuff?

PR #332, opened 2023-12-01
fix: fedimint statuses to snake case

Kodylow:


wbobeirne:
Note that this is a change to keep up to date with fedimint, not an arbitrary choice of UI to change it 👍

PR #333, opened 2023-12-01
fix: overwrite numPeers to 1 for solo

Kodylow:
in case they go back and forth on the role selector.


apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -189,7 +189,7 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
           myName,
           password,
           configs: {
-            numPeers: parseInt(numPeers, 10),
+            numPeers: isSolo ? 1 : parseInt(numPeers, 10),
```
wbobeirne, line 192: This might work, but might also be a little buggy since there's also some isSolo consideration up at line 65:
```ts
  const [numPeers, setNumPeers] = useState(
    stateNumPeers ? stateNumPeers.toString() : isSolo ? '1' : '4'
  );
```
We should do this in only one place so they don't get out of sync, either only here, or only up there by changing the logic up there to:
```ts
isSolo ? '1' : stateNumPeers ? stateNumPeers.toString() : '4'
```

wbobeirne (approved):

PR #334, opened 2023-12-01
chore: bump fedimint to release/v0.2

Kodylow:
Bumps to v0.2 branch that we're backporting fedimint side changes onto.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,19 +2,14 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05";
     flake-utils.url = "github:numtide/flake-utils";
-    fedimint = {
-      url = "github:fedimint/fedimint?rev=bbe587845935517df8a36b9b2a6b98fa4bc1e19e";
-    };
+    fedimint = { url = "github:fedimint/fedimint?branch=releases/0.2"; };
   };
   outputs = { self, nixpkgs, flake-utils, fedimint }:
     flake-utils.lib.eachDefaultSystem (system:
       let
-        pkgs = import nixpkgs {
-          inherit system;
-        };
+        pkgs = import nixpkgs { inherit system; };
         fmLib = fedimint.lib.${system};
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -2,19 +2,14 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05";
     flake-utils.url = "github:numtide/flake-utils";
-    fedimint = {
-      url = "github:fedimint/fedimint?rev=bbe587845935517df8a36b9b2a6b98fa4bc1e19e";
-    };
+    fedimint = { url = "github:fedimint/fedimint?branch=releases/0.2"; };
```
wbobeirne, line 5: Not entirely sure of nix's lock file behavior but this seems like it'll either be harder to update than using a hash or git tag, or that users may unexpectedly end up with `flake.lock` changes if someone pushes to releases/0.2. Any reason you don't want to stick with the current system of using hashes?

elsirion: You have to explicitly update the lockfile, it won't auto-update. Then it will use the most up-to-date version.

wbobeirne: Hmm, this also makes updating the README / docker compose a little harder since now it's not just a trivial find & replace. We had instructions over here which no longer work with this: https://github.com/fedimint/ui#bumping-referenced-fedimint

Kodylow: kk I can update to a hash, do we want to cut a 0.2-rc4 in line with fedimint? @elsirion is there any change to the gateway rpc from your lightning module PR?

wbobeirne: Release candidates don't track (e.g. ui doesn't have an rc-N to match a fedimint rc-N) so no need to cut an rc4 if there are no changes necessary. Updating the dockerfiles / flake.nix doesn't have any affect on the build artifacts that are attached to tags.

elsirion: I verified that the UI works with rc5.

PR #335, opened 2023-12-01
feat: no guardians card for solo

Kodylow:
Nulls guardian card for solo guardian dashboard.

### DIFF
--- a/apps/guardian-ui/src/components/GuardiansCard.tsx
+++ b/apps/guardian-ui/src/components/GuardiansCard.tsx
@@ -82,6 +82,10 @@ export const GuardiansCard: React.FC<Props> = ({ status, config }) => {
     return peerDataArray;
   }, [status, config, t]);
 
+  if (!Object.keys(status?.federation?.status_by_peer || {}).length) {
+    return null;
+  }
+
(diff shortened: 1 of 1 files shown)


apps/guardian-ui/src/components/GuardiansCard.tsx
```diff
@@ -82,6 +82,10 @@ export const GuardiansCard: React.FC<Props> = ({ status, config }) => {
     return peerDataArray;
   }, [status, config, t]);
 
+  if (!Object.keys(status?.federation?.status_by_peer || {}).length) {
+    return null;
+  }
```
wbobeirne, line 87: Any reason this isn't just `rows.length`? If the intent is to render the loading state, maybe `config && rows.length`?

wbobeirne (approved):

apps/guardian-ui/src/components/GuardiansCard.tsx
Kodylow: fixed, just checking if it's solo, using rows makes more sense

wbobeirne (approved):

PR #340, opened 2023-12-05
fix: decode string responses as json for address and deposit txid

### LINKED ISSUE #338: "View it on mempool.space" has quotes in url and is invalid
It generates a URL like:

https://mempool.space/tx/%22e9a0ca11c63b0f01e75bb47153e7af9813d99fa9afcd20e3d911b98dc9f61833%22

wbobeirne:
Closes #338. Replaces both instances of a string response with `res.json()`, not just for objects after all!


Kodylow (approved):
very nice

PR #347, opened 2024-01-07
chore: bump fedimint version to 0.2.1

Kodylow:
tested 4x and solo guardian setups + lightning gateway. I'd already fixed for working with rc-3 and was using that with clovyr the past couple weeks.


wbobeirne (commented):
Can you find and replace the old version everywhere, including the docker configs and the readme with instructions for bumping?

wbobeirne (approved):
utACK

PR #349, opened 2024-01-15
Define keyboard highlighting with open/close feature

### LINKED ISSUE #324: Collapsible fields lack accessibility
The collapsible field group headings do not allow for keyboard highlighting or opening / closing without using the mouse. They should use a `<button>` element or have tabIndex / aria fields to allow for keyboard interaction. They should also have titles for screen readers that indicate what interacting with them do.

CleanShot 2023-11-28 at 09 58 14@2x

isaack-njama:
This PR closes #324 

Adds accessibility into collapsible fields by allowing navigation using the keyboard.

Screenshot from 2024-01-15 22-15-54

image

image


packages/ui/src/FormGroupHeading.tsx
```diff
@@ -24,14 +32,17 @@ export const FormGroupHeading: React.FC<Props> = ({
       p={2}
       borderRadius={8}
       w='100%'
-      onClick={onClick}
+      tabIndex={1}
+      cursor='pointer'
```
wbobeirne, line 36: Making the cursor a pointer and giving it a tabIndex across the entire header may give a user the mistaken understanding that clicking this will open / collapse the header, but only clicking the button inside will.

packages/ui/src/FormGroupHeading.tsx
```diff
@@ -1,11 +1,19 @@
...
+  useTheme,
+  Button,
+} from '@chakra-ui/react';
 
 interface Props {
   icon?: React.FunctionComponent<React.SVGAttributes<SVGElement>>;
   title: React.ReactNode;
   onClick?: () => void;
   chevronIcon?: string;
+  tabIndex: number;
```
wbobeirne, line 16: This prop is entirely unused, and should be removed.

packages/ui/src/FormGroupHeading.tsx
```diff
@@ -24,14 +32,17 @@ export const FormGroupHeading: React.FC<Props> = ({
...
+      tabIndex={1}
+      cursor='pointer'
     >
       {icon && <Icon width='20px' height='20px' mr={2} as={icon} />}
       <Heading fontSize='md' lineHeight='20px' fontWeight='700'>
         {title}
       </Heading>
       <Spacer />
-      {chevronIcon}
+      <Button onClick={onClick} tabIndex={1} variant='ghost' color='black'>
```
wbobeirne, line 43: Buttons don't need a tabIndex, they're tabbable by default.

wbobeirne: Choosing to use the Chakra `<Button>` component added some new styling here that adds padding that wasn't there before. Maybe this is desirable, but it's a change worth considering.

wbobeirne (commented):
cc @Kodylow to take a look at how this affects the behavior (clickable only on the right) and visuals (additional padding for arrow), make sure these changes are desirable.

The issue I opened was entirely just about the accessibility of it, so I want to make sure any additional changes are considered as well.

packages/ui/src/FormGroupHeading.tsx
Kodylow: I think if it's collapsed having it be "click anywhere" is preferable

Kodylow:
echo all nits by will, and click to open should be the whole bar not just the arrow if we're updating anyway.

okjodom:
currently validating the latest iteration

packages/ui/src/FormGroupHeading.tsx
okjodom: While I think the padding unintentionally introduced by the button is pleasant, the button adds a tab stop that's redundant to the functionality of the form group header and we should not introduce it

okjodom: in case we decide to keep the button, it should also have an `onKeydown` listener

wbobeirne: `onClick` captures the typical accessibility keys for proceeding (space bar, enter etc)

wbobeirne: Sorry, I was mistaken. That is only the case if you use proper elements. Even adding accessibility attributes like `tabIndex` and `role` don't work if it's not a `button` or `a`: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#required_javascript_features

Although for what it's worth, that's a reason to get the element to use the right element instead of using a div and trying to turn it into an accessible clickable element. This would be a lot easier if it was just a `<button>`.

isaack-njama: Refactoring this to Button element seems to work just fine. No redundant functionality with tab stop, and open/close works for both onClick and keydown (Enter and Space bar) without explicitly specifying this.

isaack-njama: [refactor to button element.webm](https://github.com/fedimint/ui/assets/41724425/4b0609d2-1363-4705-b241-44e974173c34)

packages/ui/src/FormGroupHeading.tsx
```diff
@@ -16,22 +24,23 @@ export const FormGroupHeading: React.FC<Props> = ({
 }) => {
   const theme = useTheme();
   return (
-    <Flex
-      align='center'
-      justify='start'
+    <Button
       mb={3}
-      bg={theme.colors.gray[100]}
+      color={theme.colors.white[100]}
```
okjodom, line 29: try apply the right button variant and colors so we fix A11y without changing visible UI. Something like `variant='ghost'`+ `bg={theme.colors.gray[100]}` + text color modification should work

isaack-njama: Side note: Open/Close on Follower isn't that graceful. Effect of having less fields?

Kodylow:
just test with jodom, approving and merging

okjodom (approved):

Kodylow (approved):
approve

PR #351, opened 2024-01-18
Restart federation setup

### LINKED ISSUE #253: Can restart fedimint setup
Sometimes setups fail in a way that could be avoided if they were attempted again.

okjodom:
Add UX for restarting federation setup. These need design review as the UX spec doesn't include restart experience

**Validations**

- [x] Lead guardian can initiate setup restart at `SetConfiguration`, `ConnectGuardians`, `RunDKG` or `VerifyGuardians` pages
- [ ] Any follower guardian can initiate setup restart at the same pages
- [x] All guardians get notified of restart intent, and have to restart before proceeding

image
image

close #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ From root repo directory:
 
 ## Referencing Fedimint
 
-The docker containers and devimint are for specific releases or commits of `fedimint/fedimint`. At present, the reference commit-hash is `a8422b84102ab5fc768307215d5b20d807143f27`
+The docker containers and devimint are for specific releases or commits of `fedimint/fedimint`. At present, the reference commit-hash is `9ec5418ee002134188d4584cc645fe1b5ffba4f1`
 
 ### Running with local Fedimint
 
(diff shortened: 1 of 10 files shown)


okjodom:
@Kodylow , @wbobeirne do test out this restart setup PR when you can. Did we have specific designs for restart UX?

Kodylow:
The error message makes it seem like it's optional whether you should restart as well if there's another guardian who restarted. Should be more like "One of the guardians restarted the setup process".

Alternatively we could have it so only the setup leader can restart and then the message is "The Setup Leader Restarted the Ceremony". Think that would be better.

Then after I did a restart and tried again the followers all hung on this screen (shouldn't show restart here), until I refreshed. After refresh they were all fine on the validate screens:

image

Kodylow:
@wbobeirne made some updates here with jodom:

Making it so only setup leader can initiate a restart. Then it has a confirm modal for are you sure you want to restart. Then folowers get notified the setup leader restarted the process and dump back to the first setup page.

Makes it way easier to avoid the whole setup leader vs folllower ordering bug. going to get some design feedback on it but after these changes and some testing it'll be good to go.

EthnTuttle (approved):
At @Kodylow's behest.

PR #353, opened 2024-01-21
feat: Leave Federation

Kodylow:
Needs design input and I think we should use toasts for errors we've got that whole section in the top right that's blank rn.

Need design input on how to view multiple federations too, think we're at the point we can start implementing that

### DIFF
--- a/apps/gateway-ui/src/GatewayApi.ts
+++ b/apps/gateway-ui/src/GatewayApi.ts
@@ -138,6 +138,23 @@ export class GatewayApi {
     }
   };
 
+  leaveFederation = async (federationId: string): Promise<Federation> => {
+    try {
+      const res: Response = await this.post('leave-fed', {
+        federation_id: federationId,
+      });
+
+      if (res.ok) {
+        const federation: Federation = await res.json();
+        return Promise.resolve(federation);
+      }
+
+      throw responseToError(res);
+    } catch (error) {
+      return Promise.reject({ message: 'Error leaving federation', error });
+    }
+  };
+
   requestWithdrawal = async (
     federationId: string,
     msatAmountOrAll: number | string, // you can pass msat amount or 'all'
--- a/apps/gateway-ui/src/GatewayApi.ts
+++ b/apps/gateway-ui/src/GatewayApi.ts
@@ -138,18 +138,15 @@ export class GatewayApi {
     }
   };
 
-  leaveFederation = async (federationId: string): Promise<Federation> => {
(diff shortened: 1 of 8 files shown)


apps/gateway-ui/src/GatewayApi.ts
```diff
@@ -115,18 +115,15 @@ export class GatewayApi {
     }
   };
 
-  leaveFederation = async (federationId: string): Promise<Federation> => {
+  leaveFederation = async (federationId: string): Promise<void> => {
```
okjodom, line 118: nit: should have amended previous commit :p

apps/gateway-ui/src/components/FedNameCard.tsx
```diff
@@ -1,28 +1,108 @@
...
 }
 
 export const FedNameCard = React.memo(function FedNameCard({
   title,
+  federationId,
   federationName,
+  balanceMsat,
   children,
 }: FedNameCardProps) {
+  const [isOpen, setIsOpen] = React.useState(false);
```
okjodom, line 34: nit: we should use descriptive state names like `showLeaveFedDialog`

apps/gateway-ui/src/components/FedNameCard.tsx
```diff
@@ -1,28 +1,108 @@
...
+  const [errorMessage, setErrorMessage] = React.useState<string | null>(null);
```
okjodom, line 35: nit: import-style: we should just import react fns at the top

apps/gateway-ui/src/components/FedNameCard.tsx
```diff
@@ -1,28 +1,108 @@
...
+
+  const handleDisconnectFederation = async () => {
+    try {
+      if (federationId) {
+        await gateway.leaveFederation(federationId);
+        return;
+      } else {
+        throw new Error('Federation ID is null');
+      }
+    } catch (error: any) {
```
okjodom, line 47: lint: `catch (error: unknown)`

okjodom: or just leave out the type

okjodom (commented):
aside from nits and lints, this looks good to me.
great because gateway admins can now quickly leave fed

> Need design input on how to view multiple federations too, think we're at the point we can start implementing that

let's track separately: https://github.com/fedimint/ui/issues/356

Kodylow:
Updated UI:

image

image

image

okjodom (approved):

PR #357, opened 2024-01-23
fix: update local state for verify config hashes

okjodom:
fixes a state bug where `toggleConsensusPolling(false) would mean `VerifyGuardians` page doesn't activate next button even after all guardians have been verified. before this fix, the user is forced to reload page to proceed


Kodylow:
tACK

Kodylow (approved):

PR #369, opened 2024-02-07
Config json reboot

### LINKED ISSUE #224: Replace environment variables with `config.json` loaded at runtime

EthnTuttle:
Closes #224 

I am having trouble with my dev env so would appreciate a solid testing of this. It worked for me one time when things were cooperative.

### DIFF
--- a/Dockerfile.guardian-ui
+++ b/Dockerfile.guardian-ui
@@ -34,4 +34,4 @@ COPY --from=installer /app/apps/guardian-ui/build build
 COPY scripts/replace-react-env.js scripts/replace-react-env.js
 COPY scripts/write-config-from-env.js scripts/write-config-from-env.js
 RUN yarn global add serve
-CMD node scripts/write-config-from-env.js && serve -s build
+CMD node scripts/write-config-from-env.js build && serve -s build
--- a/Dockerfile.guardian-ui
+++ b/Dockerfile.guardian-ui
@@ -24,10 +24,7 @@ RUN yarn install
 COPY --from=builder /app/out/full/ .
 # TODO: Remove this copy once https://github.com/vercel/turbo/issues/3758 is fixed
 COPY ./turbo.json .
-RUN REACT_APP_FM_CONFIG_API="{{REACT_APP_FM_CONFIG_API}}" \
-  REACT_APP_TOS="{{REACT_APP_TOS}}" \
-  yarn build:guardian-ui
-
+RUN yarn build:guardian-ui
 
 # Run the built app with a minimal image
 FROM base AS runner
@@ -35,5 +32,6 @@ WORKDIR /app
 
 COPY --from=installer /app/apps/guardian-ui/build build
 COPY scripts/replace-react-env.js scripts/replace-react-env.js
+COPY scripts/write-config-from-env.js scripts/write-config-from-env.js
 RUN yarn global add serve
-CMD node scripts/replace-react-env.js build && serve -s build
+CMD node scripts/write-config-from-env.js && serve -s build
--- a/apps/guardian-ui/src/setup/FederationSetup.tsx
+++ b/apps/guardian-ui/src/setup/FederationSetup.tsx
@@ -47,13 +47,13 @@ export const FederationSetup: React.FC = () => {
     dispatch,
     api,
   } = useSetupContext();
-  const [needsTosAgreement, setNeedsTosAgreement] = useState(true);
+  const [showTosAgreement, setShowTosAgreement] = useState(false);
   const [confirmRestart, setConfirmRestart] = useState(false);
 
   useEffect(() => {
     async function getTos() {
       const tosPresent = !!(await getEnv()).tos;
-      setNeedsTosAgreement(tosPresent);
+      setShowTosAgreement(tosPresent);
     }
     getTos();
   }, []);
@@ -98,9 +98,9 @@ export const FederationSetup: React.FC = () => {
 
   switch (progress) {
     case SetupProgress.Start:
-      if (needsTosAgreement) {
+      if (showTosAgreement) {
         title = t('setup.progress.tos.title');
-        content = <TermsOfService next={() => setNeedsTosAgreement(false)} />;
+        content = <TermsOfService next={() => setShowTosAgreement(false)} />;
       } else {
         title = t('setup.progress.start.title');
         subtitle = t('setup.progress.start.subtitle');
--- a/apps/guardian-ui/src/setup/FederationSetup.tsx
+++ b/apps/guardian-ui/src/setup/FederationSetup.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import {
   Box,
   Button,
@@ -47,8 +47,16 @@ export const FederationSetup: React.FC = () => {
     dispatch,
     api,
   } = useSetupContext();
-  const [needsTosAgreement, setNeedsTosAgreement] = useState(!!getEnv().TOS);
+  const [needsTosAgreement, setNeedsTosAgreement] = useState(true);
   const [confirmRestart, setConfirmRestart] = useState(false);
+
+  useEffect(() => {
+    async function getTos() {
+      const tosPresent = !!(await getEnv()).tos;
+      setNeedsTosAgreement(tosPresent);
+    }
+    getTos();
+  }, []);
 
   const isHost = role === GuardianRole.Host;
   const isSolo = role === GuardianRole.Solo;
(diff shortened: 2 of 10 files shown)


apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -47,9 +47,17 @@ export const FederationSetup: React.FC = () => {
...
+  const [needsTosAgreement, setNeedsTosAgreement] = useState(true);
   const [confirmRestart, setConfirmRestart] = useState(false);
 
+  useEffect(() => {
+    async function getTos() {
+      const tosPresent = !!(await getEnv()).tos;
+      setNeedsTosAgreement(tosPresent);
+    }
+    getTos();
+  }, []);
```
wbobeirne, line 59: It's a shame to see a bunch of these scattered throughout the app because each one makes a separate request, and the initial rendering state of these components can be wrong. I think it would make more sense to fetch the config once, show a loader while it's fetching, and then have a React context provide it to all components from a single spot.

apps/guardian-ui/src/setupProxy.js
```diff
@@ -0,0 +1,19 @@
...
+const config_path = process.env.REACT_APP_DEV_CONFIG_PATH
+  ? process.env.REACT_APP_DEV_CONFIG_PATH
+  : 'ws://127.0.0.1';
+
+module.exports = function (app) {
+  console.log(`serving up ${config_path} for frontend`);
+  app.use(
+    '/config',
+    createProxyMiddleware({
+      target: 'http://localhost:5000/dev_config',
```
wbobeirne, line 12: Do we need a separate webserver to serve up these configs? This `app` object is a webserver that we already can now add behavior to ourselves, we can just do any logic of pulling in JSON files and serving them here.

apps/guardian-ui/src/utils/env.ts
```diff
@@ -1,15 +1,16 @@
...
-  return JSON.parse(
-    JSON.stringify({
-      FM_CONFIG_API: process.env.REACT_APP_FM_CONFIG_API,
-      TOS: process.env.REACT_APP_TOS,
-    })
-  );
+type EnvVars = {
+  fm_config_api: string;
+  tos: string;
+};
```
wbobeirne, line 4: These keys need to be typed as optional, we can't guarantee that the config will have these variables populated.

wbobeirne (changes requested):
This looks like a good start, I just think it can be done a) with only a single webserver in development, and b) with only a single request to the config.

One thing that would also be good to see in this PR is updates to the README for how to set these variables. It'd be good to write out how we're going to ask developers to set up these configs, so that we're all on the same page of what it looks like.

One last thing we'll need to do before this is complete (although could be a separate PR) is having the dockerfile generate the correct JSON file based on the env vars, instead of its current find/replace.

elsirion:
Love seeing progress here! Can't wait for being able to host the UI as a static site instead of a docker container!

apps/guardian-ui/src/setupProxy.js
```diff
@@ -0,0 +1,19 @@
+const { createProxyMiddleware } = require('http-proxy-middleware');
+
+const config_path = process.env.REACT_APP_DEV_CONFIG_PATH
+  ? process.env.REACT_APP_DEV_CONFIG_PATH
+  : 'ws://127.0.0.1';
```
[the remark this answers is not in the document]
EthnTuttle: This was a copy/pasta error on my part. Will clean up in subsequent commits.

EthnTuttle:
> This looks like a good start, I just think it can be done a) with only a single webserver in development, and b) with only a single request to the config.
> 
> One thing that would also be good to see in this PR is updates to the README for how to set these variables. It'd be good to write out how we're going to ask developers to set up these configs, so that we're all on the same page of what it looks like.
> 
> One last thing we'll need to do before this is complete (although could be a separate PR) is having the dockerfile generate the correct JSON file based on the env vars, instead of its current find/replace.

[ ] singel webserver
[ ] README update
[ ] dockerfile generated config.json from ENVVAR

Thanks for the feedback.

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -47,9 +47,17 @@ export const FederationSetup: React.FC = () => {
     dispatch,
     api,
   } = useSetupContext();
-  const [needsTosAgreement, setNeedsTosAgreement] = useState(!!getEnv().TOS);
+  const [needsTosAgreement, setNeedsTosAgreement] = useState(true);
```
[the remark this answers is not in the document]
EthnTuttle: renamed to be `show...` since it's conditional for rendering an element. updated to default to false as well.

good catch, thank you.

EthnTuttle:
@wbobeirne Just wanted to get a review of the Context changes i made for querying the config data. Will be circling back to other things as I can.

EthnTuttle:
x-post from Discord for context

I'm having a Catch22 with this config dev server. 
Intercepting the call to the config endpoint is easy enough, but how would you determine what dev config to server up? if a single webserver is serving the content, then what in that content can specify to the backedn that it's dev server #1 or w/e.
[12:17 PM]
it's all serving the same static assets so there's no anything unique to indicate this once the browser loads up the app. I suppose I could add something to the UI that is dev specific for this? Which might be desireable for using the same UI static site to switch between federations?
[12:18 PM]
Will be in voice for 30 min or so while i ponder and sort through this
[12:20 PM]
also pondering a custom error page for when the config is not found so a user can specify a new path to look in for the config?

apps/guardian-ui/src/utils/env.ts
```diff
@@ -1,15 +1,16 @@
...
+export async function getEnv() {
+  const response = await fetch('config.json');
+  if (!response.ok) {
+    throw new Error('Could not find config.json');
+  }
+  const config: EnvVars = await response.json();
+  if (config.fm_config_api === 'config api not set' || !config.fm_config_api) {
+    throw new Error('Config API not set in config.json');
+  }
+  return config;
```
wbobeirne, line 15: This type def and function are now in two places, here and `GuardianApi`

Dockerfile.guardian-ui
```diff
@@ -24,16 +24,14 @@ RUN yarn install
...
-  yarn build:guardian-ui
-
+RUN yarn build:guardian-ui
 
 # Run the built app with a minimal image
 FROM base AS runner
 WORKDIR /app
 
 COPY --from=installer /app/apps/guardian-ui/build build
 COPY scripts/replace-react-env.js scripts/replace-react-env.js
```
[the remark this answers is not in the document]
EthnTuttle: https://github.com/fedimint/ui/issues/384

apps/guardian-ui/src/utils/env.ts
EthnTuttle: Good catch. Will refactor.

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -47,9 +46,20 @@ export const FederationSetup: React.FC = () => {
     dispatch,
     api,
   } = useSetupContext();
-  const [needsTosAgreement, setNeedsTosAgreement] = useState(!!getEnv().TOS);
+  const [tosConfig, setTosConfig] = useState<{
+    showTos: boolean;
+    tos: string | undefined;
+  }>({ showTos: false, tos: undefined });
```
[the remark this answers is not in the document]
wbobeirne: Understood, my question is why are they a single piece of state as an object, instead of:
```ts
const [showTos, setShowTos] = useState(false)
const [tos, setTos] = useState<string>()
```

EthnTuttle:
Apologies folks. Lost momentum on this. No certainty I'll close on this anytime soon.

Kodylow:
can you leave todos and I'll finish it?

EthnTuttle:
> can you leave todos and I'll finish it?

Any of the outstanding conversatation basically.

- [ ] EnvVars type defined in two spots
- [ ] move config state into it's own context

That's about it from what I see. Currently this works but could be improved upon.

Dayvvo:
@EthnTuttle @Kodylow  I'd like to pick this up if you don't mind. Can you provide more context to the "move config state into it's own context" todo. Currently what I think you mean is to move tosConfig variable found at [tosConfig](https://github.com/fedimint/ui/pull/369/files#diff-7678898746f2fa77f54469178ef028a4dda9a5c6543ebac6e3cd465dadd26640:~:text=const%20%5BtosConfig,%7D)%3B) into a context state in [SetupContext.tsx](https://github.com/fedimint/ui/blob/config-json-reboot/apps/guardian-ui/src/setup/SetupContext.tsx). I'd appreciate a response. Thanks!

Kodylow:
that's my understanding from when I last talked to ethan

PR #372, opened 2024-02-08
Fix gateway api auth

Kodylow:
If you set the Gateway API in the server side environment variables it'll set it as authed for any client logging in. Removes pull from env for gateway api test password.


okjodom (approved):
lgtm

PR #373, opened 2024-02-08
fix: remove unused imports and add password documentation

Kodylow:


README.md
```diff
@@ -97,6 +97,9 @@ Replace `-p 3000:3000` with a port of your choice, `REACT_APP_FM_GATEWAY_API` wi
 
 ## Development
 
+The development password for the gateway is `thereisnosecondbest`.
```
dpc, line 100: Should we just make it `pass` as well? At least it's something one could guess if they missed it in the docs.

dpc (approved):

README.md
Kodylow: we use thereisnosecondbest for all the gateway stuff on mutinynet too so I'd suggest just keep it, 'pass' is the odd one out we just set that for all the guardians if you run the gateway dev that automatically sets the a 3/4 federation running

PR #374, opened 2024-02-12
feat: pass in query param for role

Kodylow:
Lets you pass in `role` as a query param so we can skip the first setup page, we've found it a hiccup when doing setups for fedi

### DIFF
--- a/apps/guardian-ui/src/components/RoleSelector.tsx
+++ b/apps/guardian-ui/src/components/RoleSelector.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import {
   Button,
   Icon,
@@ -30,6 +30,26 @@ export const RoleSelector = React.memo<Props>(function RoleSelector({
   const { t } = useTranslation();
   const { dispatch } = useSetupContext();
   const [role, setRole] = useState<GuardianRole>();
+
+  // If role in query params, set it
+  useEffect(() => {
+    const urlParams = new URLSearchParams(window.location.search);
+    const role = urlParams.get('role')?.toLowerCase();
+    if (role) {
+      switch (role) {
+        case 'host':
+          setRole(GuardianRole.Host);
+          break;
+        case 'follower':
+          setRole(GuardianRole.Follower);
+          break;
+        case 'solo':
+          setRole(GuardianRole.Solo);
+          break;
+      }
+    }
+  }, []);
+
   const options: RadioButtonOption<GuardianRole>[] = useMemo(
     () => [
       {


apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -87,16 +88,38 @@ export const FederationSetup: React.FC = () => {
...
   switch (progress) {
     case SetupProgress.Start:
       if (needsTosAgreement) {
         title = t('setup.progress.tos.title');
         content = <TermsOfService next={() => setNeedsTosAgreement(false)} />;
       } else {
-        title = t('setup.progress.start.title');
-        subtitle = t('setup.progress.start.subtitle');
-        content = <RoleSelector next={handleNext} />;
+        roleQueryParam = getQueryParam('role');
```
wbobeirne, line 99: Prefer using a hook for this, this isn't re-render safe (e.g. if query param changed, it would not trigger the component to re-render.)

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -87,16 +88,38 @@ export const FederationSetup: React.FC = () => {
...
+          } else if (roleQueryParam === 'solo') {
+            dispatch({
+              type: SETUP_ACTION_TYPE.SET_ROLE,
+              payload: GuardianRole.Solo,
+            });
+          } else if (roleQueryParam === 'follower') {
+            dispatch({
+              type: SETUP_ACTION_TYPE.SET_ROLE,
+              payload: GuardianRole.Follower,
+            });
```
wbobeirne, line 116: Don't call `dispatch` in the render method, render methods should be pure meaning that running them twice should produce the same result, and running a render should not modify application state. These belong in a `useEffect`.

apps/guardian-ui/src/setup/FederationSetup.tsx
wbobeirne: Eh, feel free to ignore this. This is more of a consideration when using `react-router` or something of its ilk. Since this app doesn't actually do any routing, it's a bit of a non-issue. Although moving this code to outside of the component may make it more explicitly clear that the query params are not re-render safe, and should not be used to manipulate app state.

wbobeirne (changes requested):

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -82,6 +83,22 @@ export const FederationSetup: React.FC = () => {
...
   }, [api, dispatch]);
 
+  useEffect(() => {
+    const roleQueryParam = getQueryParam('role')?.toLowerCase();
+    if (
+      roleQueryParam &&
+      progress === SetupProgress.Start &&
+      ['Host', 'Solo', 'Follower'].includes(roleQueryParam)
+    ) {
+      if (roleQueryParam) {
```
okjodom, line 93: conditional is redundant with the outer if check

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -82,6 +83,22 @@ export const FederationSetup: React.FC = () => {
       });
   }, [api, dispatch]);
 
+  useEffect(() => {
```
okjodom, line 86: nit: We can early return if `progress !== SetupProgress.Start` to avoid checking the query params.
lint: Add `progress` to the params of this useEffect?

okjodom:
Lgtm. I'll test in a bit.
PS, we can squash all commits into one

Kodylow:
squashed

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -82,6 +83,23 @@ export const FederationSetup: React.FC = () => {
...
+    if (progress !== SetupProgress.Start) {
+      return;
+    }
+    const roleQueryParam = getQueryParam('role')?.toLowerCase();
+    if (
+      roleQueryParam &&
+      progress === SetupProgress.Start &&
```
okjodom, line 93: nit: this line is now redundant :)

apps/guardian-ui/src/setup/FederationSetup.tsx
```diff
@@ -94,6 +112,7 @@ export const FederationSetup: React.FC = () => {
         title = t('setup.progress.tos.title');
         content = <TermsOfService next={() => setNeedsTosAgreement(false)} />;
       } else {
+        // Removed the roleQueryParam handling from here as it's moved to useEffect
```
okjodom, line 115: we can delete the comment

okjodom:
@Kodylow when I was testing this, I found some bugs (tagged) but I actually took a step back wondering if we actually need to skip the first page because of query param. Might it be better to pre-select the role from the value of query param, but let the guardian click next to proceed?

Kodylow:
I agree that's better, then people acknowledge their role too. will update.

PR #377, opened 2024-02-16
feat: fix nix flake

justinmoon:


okjodom:
what is broken on the current flake?

justinmoon:
I believe this depends on https://github.com/fedimint/fedimint/pull/4349

Kodylow:
should we use that same working commit or a different one? and do we actually need to fix it on the release, it's just a dev env issue?

justinmoon:
We don't need to fix `v0.2` if you we won't need to change it at all. Perhaps it's good enough as-is. It seemed like you wanted to change something on `v0.2` but maybe I was mistaken.

Kodylow:
no just needed to change to to get the dev environment working

PR #383, opened 2024-02-23
Fix: flake.nix

Kodylow:
should use fedimint not dpc fork


elsirion (approved):
Doesn't matter as long as the hash is identical, but probably good to do.

Issue #385, opened 2024-03-01
Refreshing the setup leader kicks off DKG as solo or with any connected guardians


Kodylow:
Bug reported by bullmoose in fedimint discord and I replicated locally. If you hit refresh on the setup leader while you're waiting for the other guardians to connect to its websocket link then on refresh it loads as numPeers = 1 (just itself) so calls next() immediately. have to persist guardians and some other state through the refresh, and should also confirm that it's selected Solo not just that numPeers = 1

PR #386, opened 2024-03-01
fix: set numPeers correctly

Kodylow:
fixes https://github.com/fedimint/ui/issues/385

We weren't properly loading the state from localStorage, and we were setting numPeers from connected Peers not from stateNumPeers. This lets the setup leader refresh and maintain state and connections to peers, and doesn't kick to DKG immediately on refresh.

### DIFF
--- a/apps/guardian-ui/src/components/SetConfiguration.tsx
+++ b/apps/guardian-ui/src/components/SetConfiguration.tsx
@@ -41,6 +41,8 @@ interface Props {
   next: () => void;
 }
 
+const MIN_BFT_NUM_PEERS = '4';
+
 export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
   const { t } = useTranslation();
   const {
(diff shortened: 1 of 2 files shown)


apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -75,6 +72,9 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
   });
   const [mintAmounts, setMintAmounts] = useState<number[]>([]);
   const [error, setError] = useState<string>();
+  const [numPeers, setNumPeers] = useState(
+    stateNumPeers ? stateNumPeers.toString() : isSolo ? '1' : '4'
```
EthnTuttle, line 76: Why are we hard coding four? Can we define a const or function for this instead?

Kodylow: the hardcode is what it starts with for if they're a host, so in their meta configuration it starts at 4

flake.lock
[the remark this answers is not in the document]
Kodylow: was updated in a prev manual PR, missed committing the lock with it. just switches from /dpc to /fedimint

apps/guardian-ui/src/setup/SetupContext.tsx
```diff
@@ -23,19 +23,7 @@ const LOCAL_STORAGE_SETUP_KEY = 'setup-guardian-ui-state';
  * Creates the initial state using loaded state from local storage.
  */
 function makeInitialState(loadFromStorage = true): SetupState {
```
okjodom, line 25: Hey @Kodylow , this is the actual fix to the bug! Nice fix. Could we just scope the PR to changes in this file and we land it?

okjodom (commented):
tACK. Looks good, but I wish it was just a super small change that only includes the changes in `SetupContext.tsx`. We can append the other changes as extra opportunistic commits to clean up the code.

PR #388, opened 2024-03-05
feat: download guardian backup

Kodylow:
Bumps flake to most recent fedimint master commit to get the download guardian config endpoint and adds flow to download backup: button next to Federation Config + modal to acknowledge they're downloading private key material

### DIFF
--- a/apps/guardian-ui/src/languages/en.json
+++ b/apps/guardian-ui/src/languages/en.json
@@ -59,7 +59,7 @@
       "downloadBackup": {
         "button": "Download Backup",
         "warningTitle": "Warning",
-        "warningText": "The backup contains private keys and secret material for the federation guardian and must be kept secure.\n Recovery using this backup requires your admin password!",
+        "warningText": "The backup contains private keys and secret material for the federation guardian and must be kept secure. Recovery using this backup requires your admin password!",
         "cancelButton": "Cancel",
         "acknowledgeButton": "Acknowledge & Download"
       }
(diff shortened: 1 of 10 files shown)


apps/guardian-ui/src/languages/en.json
```diff
@@ -55,7 +55,14 @@
       "no-gateways-info-description": "Lightning node operators can connect to your federation to provide Lightning Network interoperability. Once connected, they will appear here."
     },
     "config": {
-      "label": "Federation Config"
+      "label": "Federation Config",
+      "downloadBackup": {
+        "button": "Download Backup",
+        "warningTitle": "Warning",
+        "warningText": "The backup contains private keys and secret material for the federation guardian and must be kept secure.\n Recovery using this backup requires your admin password!",
```
otech47, line 62: didn't test this but from my understanding of the `useTranslation` hook in other projects, line breaks won't work out of the box like this. might need to render it in a `<Trans>` component for this to do what you expect.

otech47 (approved):

apps/guardian-ui/src/languages/en.json
Kodylow: just removed linebreak, isn't necessary.

Kodylow:
moved button and model into new DownloadBackup.tsx and moved it to the bottom of the screen, design feedback was that putting it on the Federation Config meant it look like that was the thing they were backing up not the whole guardian dir 
image

otech47 (approved):

justinmoon (approved):

PR #392, opened 2024-03-16
chore: bump fedimint to meta module

Kodylow:
Bumps fedimint version in nix flake to commit including meta module so I can start on the guardian UI for meta updates


okjodom (approved):
🔥

Issue #393, opened 2024-03-16
Autogenerate guardian password


isaack-njama:
Related to https://github.com/fedimint/fedimint/issues/4364

PR #394, opened 2024-03-16
feat: define logic to autogenerate guardian password

isaack-njama:
Addresses #393 

This logic generates a 16-character password that guardians can copy and backup and present during password check.

Question: At what point should password check happen? 

**Updated user flow**:
image

[Screencast from 02-04-2024 10:41:39 ASUBUHI.webm](https://github.com/fedimint/ui/assets/41724425/143120c3-2d68-40e1-a727-c5b1f90e321b)

### DIFF
--- a/apps/guardian-ui/src/components/SetConfiguration.tsx
+++ b/apps/guardian-ui/src/components/SetConfiguration.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import {
+  useClipboard,
   Flex,
   FormControl,
   FormLabel,
@@ -10,7 +11,6 @@ import {
   Button,
   Text,
   useTheme,
-  FormErrorMessage,
 } from '@chakra-ui/react';
 import {
   BitcoinRpc,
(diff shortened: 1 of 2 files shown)


Kodylow:
We should also add an acknowledgement checkbox that the user has copied down the secure password

isaack-njama:
> We should also add an acknowledgement checkbox that the user has copied down the secure password

Updated the PR description with screenshot and short video of UI with most recent changes.

The idea is to not allow a guardian to proceed with federation configuration until they acknowledge that the password has been copied and securely backed up.

isaack-njama:
My observation is that the password is the same in a session for all guardians (leader, follower, and solo) during the `Federation details` phase of setup.

Would it be necessary to refresh the password after a certain amount of time (e.g. 10 / 20 seconds)?

Kodylow:
> My observation is that the password is the same in a session for all guardians (leader, follower, and solo) during the Federation details phase of setup.

What do you mean here? the password is supposed to be the same on the same fedimint guardian instance regardless of what role they choose in the setup ceremony?

isaack-njama:
> > My observation is that the password is the same in a session for all guardians (leader, follower, and solo) during the Federation details phase of setup.
> 
> What do you mean here? the password is supposed to be the same on the same fedimint guardian instance regardless of what role they choose in the setup ceremony?

Makes sense. It's just a concern I had, thought I should mention it.

Kodylow (approved):
lgtm

PR #399, opened 2024-03-25
feat: show federation name in gateway ui

### LINKED ISSUE #368: Add back federation name to gateway api
Not sure where the call got made to remove it but functionally for users we need to add that back in, just identifying based on federationId is impossible to manage when you're connected to multiple feds

okjodom:
fixes #368 
image


Kodylow (approved):
think federation name is optional in meta otherwise good to go

okjodom:
> think federation name is optional in meta otherwise good to go

ack. I made federation name in `BalanceCardProps` nullable. The experience falls back to the federation id as name

PR #401, opened 2024-03-26
feat: api for fetching federation id

okjodom:
adds `GuardianApi` method for fetching federation id. Enabled by https://github.com/fedimint/fedimint/pull/4576


Kodylow (approved):
ack

PR #403, opened 2024-03-27
chore: bump fm to v0.3.0-rc.3

okjodom:
- commit # c32bbe8ed8d95ae420b5879e92dda6ce48f5c914

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,7 +3,7 @@
     flake-utils.url = "github:numtide/flake-utils";
     fedimint = {
       url =
-        "github:fedimint/fedimint?rev=c51ea0c85576008dcbf9d94e2fe994dfa34bca61";
+        "github:fedimint/fedimint?rev=c32bbe8ed8d95ae420b5879e92dda6ce48f5c914";
     };
   };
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -3,7 +3,7 @@
     flake-utils.url = "github:numtide/flake-utils";
     fedimint = {
       url =
-        "github:fedimint/fedimint?rev=c51ea0c85576008dcbf9d94e2fe994dfa34bca61";
```
maan2003, line 6: are you sure? 
rc-3 is https://github.com/fedimint/fedimint/commit/c32bbe8ed8d95ae420b5879e92dda6ce48f5c914

maan2003 (changes requested):

flake.nix
maan2003: the commit you use doesn't have https://github.com/fedimint/fedimint/pull/4726
which was entire point of rc.3

okjodom: Looks like I picked the parent to rc.3 Correcting!

maan2003 (approved):

PR #404, opened 2024-03-27
refactor: download backup card

okjodom:
Just makes the download backup experience more aligned with existing cards on page
image


Kodylow:
I have a refactor of this within "danger zone" as part of my guardian auth PR that I think is a better approach given how catastrophic it could be if they fuckup backing up and lose it or whatever

okjodom:
> I have a refactor of this within "danger zone" as part of my guardian auth PR that I think is a better approach given how catastrophic it could be if they fuckup backing up and lose it or whatever

that makes more sense

Issue #397, opened 2024-03-22
Follower guardians cannot submit config gen params


okjodom:
After connecting to the lead guardian, the RPC call by followers to submit their own config gen params fails with the error below
![image](https://github.com/fedimint/ui/assets/11217077/bdb648c1-97fe-45d3-8d10-58c9a2cb0cc9)

Replicated on master branch @https://github.com/fedimint/ui/commit/b9bd211e3b69a89e85a18ebdc98bcacde785296e

okjodom:
I've been investigating this for two hours now. I noted this bug does not repro when we run setup UI on local FM with `env DEVIMINT_BIN=$(realpath ../fedimint/target/debug) yarn nix-guardian`

okjodom:
When followers request config gen params from leader, they get an error, with the leader server reporting the following error
```
2024-03-27T17:26:25.730039Z  WARN net::api: api request error path="consensus_config_gen_params" error=ApiError { code: 400, message: "Config params were not set on this guardian"
 }
 ```

okjodom:
I suspect that after https://github.com/fedimint/fedimint/pull/4520 , rpc by peers trying to get consensus_config_gen_params from lead peer server fails with the above error.

Do you know what could be goign wrong, @dpc

Kodylow:
Working on this rn, also noticed that setup leader adds the peer as pending even though it fails on the follower with missing field 'consensus'

![image](https://github.com/fedimint/ui/assets/74332828/82a1d127-2888-46cf-9a05-166a59bf8acd)

<img width="496" alt="image" src="https://github.com/fedimint/ui/assets/74332828/3ad13a48-a605-46c1-81d5-34899b748402">

Kodylow:
The "consensus" value here in `ConfigGenModuleParams` is required to parse the submitted followers config. but the follower config only submits "local"

![image](https://github.com/fedimint/ui/assets/74332828/1301882e-33f5-4a27-aea3-11178fb10ab3)

<img width="368" alt="image" src="https://github.com/fedimint/ui/assets/74332828/b64d6883-69be-4162-a915-60a377a1af11">

Kodylow:
Introduced by this change by @dpc , going to try mirroring the local values into consensus but not sure what the difference is between them functionally for the setup. We have the followers submitting just local and the leader submits both local and consensus currently which is causing the error

[...]

PR #406, opened 2024-03-30
fix: required config gen local and consensus + fix password runtime error

Kodylow:
Fixes https://github.com/fedimint/ui/issues/397

Consensus and Local are now required for submitting module params in setup so we don't filter them out anymore for the follower, plus me and eric ran into a runtime issue where the password was nulled on refresh that this fixes by doing a !!password check

### DIFF
--- a/apps/guardian-ui/src/setup/SetupContext.tsx
+++ b/apps/guardian-ui/src/setup/SetupContext.tsx
@@ -285,6 +285,7 @@ export const SetupContextProvider: React.FC<SetupContextProviderProps> = ({
           // Followers set their own connection name, and hosts server URL to connect to.
           await api.setConfigGenConnections(myName, configs.hostServerUrl);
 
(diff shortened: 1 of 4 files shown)


elsirion (approved):

apps/guardian-ui/src/setup/SetupContext.tsx
```diff
@@ -285,6 +285,7 @@ export const SetupContextProvider: React.FC<SetupContextProviderProps> = ({
           // Followers set their own connection name, and hosts server URL to connect to.
           await api.setConfigGenConnections(myName, configs.hostServerUrl);
 
+          console.log('configs', configs);
```
alexlwn123, line 288: Nit: cleanup console logs?

alexlwn123 (commented):
one nit, otherwise lgtm

PR #407, opened 2024-04-03
Random Guardian Names

Kodylow:
This just sets a random guardian name for the user instead of letting them pick one. 

Until we have something like a nostr login where you can testify to being a guardian the names here are useless and cause a bunch of headaches during setup if users don't want to correlate that they're also guardians.

### DIFF
--- a/apps/guardian-ui/src/languages/en.json
+++ b/apps/guardian-ui/src/languages/en.json
@@ -96,7 +96,8 @@
   },
   "set-config": {
     "guardian-name": "Guardian name",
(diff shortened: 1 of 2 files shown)


okjodom:
This makes most sense in dev scenarios where one person is pretending as several guardians. In scenarios with several actual guardians, I imagine random names takes away "chance of agency" for the guardians, and probably makes it more painful to delete the pre-filled random name, to write their own names

okjodom (approved):
makes sense. Ship it

PR #412, opened 2024-04-16
chore: update Readme instructions for mprocs setup

Dayvvo:
The setup instructions for running fedmint UI in development mode using mprocs do not include the instruction to run 'yarn install'

The readme was updated to to include that step


elsirion (approved):
I remember running into that problem, I wonder if the nix dev shell automates it these days.

PR #424, opened 2024-05-17
feat: Balance Sheet

Kodylow:
Adds a balance sheet view of all modules.

image

## Summary by CodeRabbit

- **New Features**
  - Introduced a unit selection feature in the `BalanceCard` component to switch between `msats`, `sats`, and `btc`.
  - Added a new `BalanceTable` component to display balance information based on the selected unit.

- **Enhancements**
  - Updated the layout of the `FederationAdmin` and `FederationInfoCard` components for improved visual presentation.
  - Changed the label "Balance" to "Balance Sheet" for clearer terminology.

- **Bug Fixes**
  - Adjusted the equity display in the `BalanceTable` to ensure accurate representation of total equity.

- **Documentation**
  - Added a new `formatValue` function to handle unit conversions.

- **Refactor**
  - Reorganized component imports and state management in the `BalanceCard` component for better performance and readability.

### DIFF
--- a/apps/guardian-ui/src/components/BalanceTable.tsx
+++ b/apps/guardian-ui/src/components/BalanceTable.tsx
@@ -1,9 +1,22 @@
 import React from 'react';
-import { Table, Tbody, Td, Th, Thead, Tr, Icon, Flex } from '@chakra-ui/react';
+import {
+  Table,
+  Tbody,
+  Td,
+  Th,
+  Thead,
+  Tr,
+  Icon,
+  Flex,
+  Box,
+  Text,
+  Grid,
+  GridItem,
+} from '@chakra-ui/react';
 import { ReactComponent as CheckIcon } from '../assets/svgs/check-circle.svg';
 import { ReactComponent as CloseIcon } from '../assets/svgs/x-circle.svg';
 import { AuditSummary, ModuleKind, MSats } from '@fedimint/types';
-import { formatMsatsToBtc } from '@fedimint/utils';
+import { formatValue } from '@fedimint/utils';
 
 interface BalanceTableProps {
   auditSummary: AuditSummary;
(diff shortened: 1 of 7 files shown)


Kodylow:
Added a filter to now put the "unknown" module in the balance sheet, that's just there to test if it'll break clients and it doesn't have any assets/liabilities

elsirion:
That `unknown` module would be confusing to users imo, but it's only present in test environments afaik, so this should be fine.

apps/guardian-ui/src/components/BalanceTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+  unit: 'msats' | 'sats' | 'btc';
+}
+
+export const BalanceTable: React.FC<BalanceTableProps> = ({
+  auditSummary,
+  unit,
+}) => {
+  const moduleSummaries = Object.entries(auditSummary.module_summaries).filter(
+    ([, module]) => module.kind !== ModuleKind.Unknown
+  );
```
elsirion, line 29: Should we filter the unknown module? It's confusing, but only there in devimint afaik.

apps/guardian-ui/src/components/BalanceTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const totalAssets = moduleSummaries.reduce((sum, [, module]) => {
+    return sum + (module.net_assets > 0 ? module.net_assets : 0);
+  }, 0) as MSats;
+  const totalLiabilities = moduleSummaries.reduce((sum, [, module]) => {
+    return sum + (module.net_assets < 0 ? module.net_assets : 0);
+  }, 0) as MSats;
+  const isBalanced = totalAssets + totalLiabilities === 0;
```
elsirion, line 37: This will be false for any federation charging fees. I think as long as assets > liabilities we should say it's balanced.

Kodylow: We should probably have a way of specifically separating out the fees either into an "equity" column or a way to get the specific amounts collected in fees from the different modules so that we can still display that the outstanding liabilities match the outstanding reserves exactly.

elsirion: Modules don't collect fees themselves and we don't record fees anywhere. All that happens is that the federation slowly becomes overfunded. We discussed different approaches, but this didn't seem like a hug problem so far. If you want you can just display any overfunding as a fees entry on the liabilities side.

Kodylow:
Okay added an Equity (Guardian Fees) column and adjusted the check to "Is Full Reserve" checking that assets > liabilities and equity is > 0

Kodylow:
Actually that took up a lot of real estate, moved into separate cards for Full reserve and fees calculations, if there's no fees it just doesn't show that card:

image

packages/types/src/modules.ts
```diff
@@ -5,6 +5,7 @@ export enum ModuleKind {
   Mint = 'mint',
   Wallet = 'wallet',
   Meta = 'meta',
+  Unknown = 'unknown',
```
elsirion, line 8: I worry a bit about this enum, does this mean we will encounter errors when encountering other modules than the ones listed here?

elsirion (approved):
Not a thorough review

packages/types/src/modules.ts
Kodylow: Going to land this and open as a separate issue for unknown module handling

PR #443, opened 2024-07-01
chore: larger charset passwords

Kodylow:
adds caps and special characters for password generation, entropy increase from 82.7184 -> 98.72 bits , since we're making the generatePassword function the required firststep we should force best practices.

## Summary by CodeRabbit

- **New Features**
  - Enhanced password generation function to include uppercase letters and special characters, increasing password complexity and security.

### DIFF
--- a/apps/guardian-ui/src/utils/index.ts
+++ b/apps/guardian-ui/src/utils/index.ts
@@ -24,7 +24,8 @@ export const generatePassword = () => {
   };
 
   const passwordLength = 16;
-  const charSet = 'abcdefghjkmnpqrstuvwxyz0123456789';
+  const charSet =
+    'abcdefghjkmnpqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ0123456789!@#$%^&*()';
 
   let adminPassword = '';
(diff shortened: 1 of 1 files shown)


apps/guardian-ui/src/utils/index.ts
```diff
@@ -24,7 +24,8 @@ export const generatePassword = () => {
   };
 
   const passwordLength = 16;
-  const charSet = 'abcdefghjkmnpqrstuvwxyz0123456789';
+  const charSet =
+    'abcdefghjkmnpqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ0123456789!@#$%^&*()';
```
alexlwn123, line 28: @Kodylow unrelated to this pr - I'm not an expert on this stuff, but the random generation code feels a little sketch. Might want to double check with @elsirion on the password generation strategy

alexlwn123 (approved):

apps/guardian-ui/src/utils/index.ts
Kodylow: he okayed it last time, was 80 bit entropy when optional but since we're requiring now we should do the increased charset

PR #464, opened 2024-08-05
fix: only allow bft numbers

Kodylow:
Only allows BFT numbers for federation setups


apps/guardian-ui/src/components/setup/screens/setConfiguration/FederationSettingsForm.tsx
```diff
@@ -24,6 +24,8 @@ export const FederationSettingsForm: React.FC<FederationSettingsFormProps> = ({
 }) => {
   const { t } = useTranslation();
 
+  const bftNumbers = [4, 7, 10, 13, 16, 19];
```
alexlwn123, line 27: nit: this should probably go into a `util/constants` file... 

Also should add `as const` at the end to make it readonly

alexlwn123 (approved):
One comment, otherwise lgtm

alexlwn123 (approved):

PR #502, opened 2024-08-16
chore: require 5 blocks for mainnet

Kodylow:
Unsafe to run with less on mainnet, disables the option to go under 5

## Summary by CodeRabbit

- **New Features**
	- Enhanced Bitcoin settings form to display dynamic helper text based on the selected network.
	- Added context-sensitive information regarding block confirmations for Mainnet usage.
  
- **Improvements**
	- Updated minimum block confirmation values based on the selected network, improving the form's responsiveness and user experience.


apps/guardian-ui/src/languages/en.json
```diff
@@ -227,6 +227,7 @@
     "bitcoin-settings": "Bitcoin settings",
     "block-confirmations": "Additional Block Confirmations",
     "block-confirmations-help": "Fedimint Guardians trail the blockchain tip by several confirmations to avoid block reorganizations.",
+    "block-confirmations-help-mainnet": "Running on Mainnet requires at least 5 additional confirmations.",
```
elsirion, line 230: Is what you choose in the admin UI the finality delay or the min confirmations? "Additional confirmations" sounds weird.

elsirion (approved):
Help text could be better, but requiring 5 or 6 confirmations seems reasonable (although I'll surely work around it at some point because I'm reckless 🤣)

apps/guardian-ui/src/languages/en.json
Kodylow: It's 1 confirmation + whatever you set in the UI that gets set, that's why I say "additional" as in "beyond 1"

elsirion: I guess we could just talk about confirmations and then transform to finality delay? On the server side it made some sense when looking from consensus perspective, but for users this seems confusing.

bradleystachurski: > It's 1 confirmation + whatever you set in the UI that gets set, that's why I say "additional" as in "beyond 1"

Should this comment say more than?

https://github.com/fedimint/ui/blob/41e5b309d888992228d2488e9be0dae83c14623f/apps/guardian-ui/src/components/setup/screens/setConfiguration/SetConfiguration.tsx#L167

PR #590, opened 2025-01-07
fix: merge locales from legacy gateway app into router

alexlwn123:
* Creates a script to merge legacy locale files into the main "Router" locales
* Runs the script and removed old files


alexlwn123:
Realized the script didn't merge the files correctly... fixing

PR #602, opened 2025-01-24
Improve Connect Service Handler

tsmith123:
This PR helps to prevent empty or invalid urls from being added in modal. Currently users can click Confirm regardless of what's in the input box.

Screenshot 2025-01-24 at 11 00 39

- Moves `getServiceType` to helper dir at a higher level so can be used in other areas of app if required
- Adds unit tests for new `getServiceType`
- Adds better typing for Guardian & Gateway types
- Adds logic to `onClick` handler in modal 

## Summary by CodeRabbit

- **New Features**
	- Added a new function to determine service type based on URL configuration
	- Enhanced validation for service URL input in Connect Service Modal

- **Refactor**
	- Moved `getServiceType` function from `utils.ts` to `service.ts`
	- Updated import path for `getServiceType`

- **Tests**
	- Added comprehensive test suite for `getServiceType` function to validate different URL scenarios

### DIFF
--- a/apps/router/src/helpers/service.test.ts
+++ b/apps/router/src/helpers/service.test.ts
@@ -1,0 +1,35 @@
+import { getServiceType } from './service';
+
+describe('helpers/service', () => {
+  describe('when empty string is provided', () => {
+    it('should return null', () => {
+      const result = getServiceType('');
+
+      expect(result).toBe(null);
+    });
+  });
+
+  describe('when invalid url is provided', () => {
+    it('should return null', () => {
+      const result = getServiceType('invalid-url');
+
+      expect(result).toBe(null);
+    });
+  });
+
+  describe('when valid guardian url is provided', () => {
+    it('should return guardian', () => {
+      const result = getServiceType('wss://test-url.com');
+
+      expect(result).toBe('guardian');
+    });
+  });
+
+  describe('when valid gateway url is provided', () => {
+    it('should return gateway', () => {
+      const result = getServiceType('https://test-url.com');
+
+      expect(result).toBe('gateway');
+    });
+  });
+});
--- a/apps/router/src/types/index.tsx
+++ b/apps/router/src/types/index.tsx
@@ -1,3 +1,9 @@
+export enum Service {
+  Guardian = 'guardian',
+  Gateway = 'gateway',
+}
+
+// Better to use enum above
 export type ServiceType = 'guardian' | 'gateway';
 
 export interface ServiceConfig {
--- a/apps/router/src/types/index.tsx
+++ b/apps/router/src/types/index.tsx
@@ -1,9 +1,3 @@
-export enum Service {
-  Guardian = 'guardian',
-  Gateway = 'gateway',
-}
-
-// Better to use enum above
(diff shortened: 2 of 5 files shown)


apps/router/src/helpers/service.test.ts
alexlwn123, line 1: tiny nit: might want to rename `helpers` to `utils` just to keep things consistent

apps/router/src/types/index.tsx
```diff
@@ -1,3 +1,9 @@
+export enum Service {
+  Guardian = 'guardian',
+  Gateway = 'gateway',
+}
+
+// Better to use enum above
```
alexlwn123, line 6: What's the rationale to use an enum over a type?

alexlwn123 (approved):
One nit. Otherwise LGTM

apps/router/src/helpers/service.test.ts
tsmith123: I did consider this. My thinking is I've always used "utils" for functions that could be used in any project. Their functionality isn't specific to the project they're used in. They're generic. Helpers on the other hand are specific to the project so in this case, establishing whether a url is a guardian or not could only be used here.

Hoping ultimately to have helpers and utils at the src level as there's lots of different util dirs in the project at different levels.

That said, happy to call it all utils if my conventions are a bit "weird" 😆

apps/router/src/types/index.tsx
tsmith123: Happy to use types - I just don't like seeing string comparisons throughout the app like this `serviceType === 'guardian'` - this creates scope for typos and is arguable less readable.

I feel this is better `serviceType === Service.Guardian`

apps/router/src/helpers/service.test.ts
alexlwn123: ahh, that makes sense! Fine to keep it as is 😄

apps/router/src/types/index.tsx
alexlwn123: gotcha, I appreciate your point regarding readability. But I think current TS best practices suggest avoiding enums for a few reasons, including:

1. Since Enums are compiled to JavaScript objects in runtime, they won't work with some TS type transformations. Union types let us create derived types within the TS type system like: 
```ts
type Service = { serviceType: 'guardian'; peers: Peer[]; }  // guardian-only types
             | { serviceType: 'gateway'; ... } // gateway-only types

type Guardian = Extract<Service, { serviceType: 'guardian' }> 
// { serviceType: 'guardian'; peers: Peer[] }

// This wouldn't work if `serviceType` was an enum.
```

2. Regarding preventing typos, union types provide autocomplete in editors, and type errors can be caught with linters. If we don't like the inline string comparison, maybe we can use an `isGuardian` helper function?

---

This being said, if you disagree I can live the enum. Your call. 

(also, I realize we currently have a bunch of enums all over the repo. 💀💀💀 Personally hoping we can correct that over time.)

tsmith123: Some good points. Let's stick with what we had. I will update. Thanks for the info 😉

alexlwn123 (approved):
Thanks for the changes!

PR #606, opened 2025-01-31
Update Turbo

tsmith123:
This PR updates the Turbo package which will make new features available to us when we need them.

- Update Turbo to v2
- Update GitHub workflow file to run tests in build phase
- Tweak Dockerfile


Dockerfile
```diff
@@ -3,27 +3,26 @@ FROM node:lts-alpine AS base
 
 # Assemble a pruned version of the repo containing only what's needed for router
 FROM base AS builder
-RUN apk add --no-cache libc6-compat
 RUN apk update
+RUN apk add --no-cache libc6-compat
```
otech47, line 7: why does the order matter here exactly?

tsmith123: I don't think it does tbh. Just bringing it in line with latest Dockerfie example from Turbo docs.

alexlwn123 (approved):
Works great! Great improvement

PR #608, opened 2025-02-01
chore: self-host PR previews

elsirion:
I built a [self-hosted vercel replacement](https://github.com/elsirion/sspd) (well, for the preview part, we don't need the remaining cloud fancyness), see what it's supposed to look like here: https://github.com/elsirion/fedimint-observer/pull/57#issuecomment-2629152348

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,43 @@
+name: Preview
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


tsmith123:
What's the plan with this PR @elsirion?

elsirion:
I think vercel previes became public somehow? That would mean less need for replacing it. I'd need to find some time to tinker with GH Actions to make it work :/

PR #610, opened 2025-02-03
wTom/remove config file

tsmith123:
This PR removes config file logic as it doesn't appear to be required (though I may be overlooking something). I found quite a bit of code that duplicates other code so I was able to remove considerable amounts.

- Remove duplicate logic from `AppContext`
- Add `useAppInit` hook with tests for initialisation and remove logic from `App.tsx`
- Modify flow of app so that user will always see the Homepage list of Guardians/Gateways (my thinking is that this will lead to a more consistent experience for all users)
- Remove various that aren't required
- Update existing tests
- Minor tweaks


alexlwn123:
Nice. I'm a big fan. Did a cursory review and tested. everything looks like it's working

alexlwn123:
Love the Red
image

tsmith123:
Yeah me too. It feels good to delete. 😂

tsmith123:
I will run final checks and take it out of draft. Thanks for checking.

PR #611, opened 2025-02-05
Update Modals (and Prevent New Gateway Creation)

tsmith123:
This PR makes the modals on the Home page more consistent.

Also, due to Gateway UI not being production ready, it offers a way to prevent new Gateways being added. (This is only a suggestion so please let me know if this approach is too heavy handed and you prefer a different approach). 

- Move action buttons on add, edit and delete service modals to the right hand side
- Make spacing consistent across add, edit and delete service modals
- Adds translation for new error message
- Adds translation for existing error message

**New Delete Service Modal**
Screenshot 2025-02-05 at 19 59 51

**New Edit Service Modal**
Screenshot 2025-02-05 at 19 41 35

**New Add Service Modal**
Screenshot 2025-02-05 at 19 41 54

**Add Service Modal (with error message)**
Screenshot 2025-02-05 at 20 02 13

### DIFF
--- a/apps/router/src/languages/en.json
+++ b/apps/router/src/languages/en.json
@@ -219,7 +219,9 @@
       "helper-text": "Will be wss:// for guardian or https:// for gateway",
       "service-type-label": "Service Type",
       "service-name-label": "Service Name",
-      "sync-status-label": "Sync Status"
+      "sync-status-label": "Sync Status",
+      "service-exists-error": "Service already exists",
+      "no-gateways-error": "Only Guardians can be added right now"
     },
     "remove-service-modal": {
       "title": "Remove {{type}}",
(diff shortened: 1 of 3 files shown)


elsirion:
Sounds good enough for now, let's work on GW UI once we got a stable release again

elsirion:
As discussed on the call, in the short term the Guardian UI is more important indeed, but if we want more people to run gateways eventually we need to visualize the complex data associated with them. In that sense a good UI is even more important for Gateways than Guardians, there's just much more information.

I also see community-run Gateways as a crucial component for widespread Fedimint adoption. Professional Gateway providers will always need to take custody of some user funds since they don't have a trust relationship to the federation. Having community members with a trust relationship run Gateways fixes this. We are already working towards that by integrating LDK, which should ultimately allow zero-management Gateways by connecting to an LSP.

tsmith123:
Thanks for the input. What I will do is restore the original logic re: gateways on this PR to keep it just abut the modals. Then I will create a new PR to prevent a user accessing the Gateway UI as a temporary measure but I will tackle it on the password page as this current approach would be bypassed unwittingly by providing the Gateway Url Env Var anwyay.

PR #614, opened 2025-02-07
feat: disable DockerHub upload except for tags

alexlwn123:
Our dockerhub is far too noisy. [see here](https://hub.docker.com/r/fedimintui/fedimint-ui/tags)

Let's only upload tagged commits/releases to dockerhub. 

cc @tsmith123

### DIFF
--- a/.github/workflows/fedimint-ui-docker.yml
+++ b/.github/workflows/fedimint-ui-docker.yml
@@ -25,14 +25,13 @@ jobs:
           images: |
             fedimintui/fedimint-ui
           tags: |
-            type=ref,event=branch
-            type=ref,event=pr
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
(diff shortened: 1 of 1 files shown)


tsmith123:
Sounds like a plan. I've been looking into it already.

elsirion (approved):
Let's go I'd say!

tsmith123:
Just a thought but would it be better to change the "trigger" for this job rather than using the conditionals? i.e. "on release"

alexlwn123:
> Just a thought but would it be better to change the "trigger" for this job rather than using the conditionals? i.e. "on release"

Was thinking we'd still want try to build the containers on every PR, just not push them to dockerhub

alexlwn123:
Taking another look, I realize that's not actually the case... implementing your suggestion now

tsmith123:
So branches get built using the `ci` workflow and nothing has changed there. My suggestion was really to clean the docker workflow file up a little to save using the conditionals. Feel free to merge if you want now - I can always add some updates further down the line.

PR #619, opened 2025-02-17
Homepage Redesign

tsmith123:
This PR introduces a modernised Homepage and steers the user away from managing multiple guardians and gateways.

- Modern and fresh looking homepage
- Allows just one service rather than multiple
- Adds sticky footer and uses icons for social links
- Removes modals for adding, editing and deleting and replaced with single input box
- Usability extras including ESC key and auto focus
- Prevents Gateway urls (whilst Gateway UI not ready for production)
- Considerable code clean up across multiple files
- Adds more unit tests
- Removes unused keys from all language files
- Fully responsive

(Note: hero images are from Undraw and are free to use in any project as per their licence agreement https://undraw.co/license)

**New Desktop/Laptop Layout**
new-hp-1

**New Responsive Layout**
new-hp-2

**Alternative Hero Image**
hero-2


joschisan (approved):

apps/router/src/components/Footer.tsx
```diff
@@ -24,26 +24,39 @@ export const Footer = () => {
...
-        <CustomLink title={version.display} href={version.url} />
-        <CustomLink title='Discord' href='https://chat.fedimint.org/' />
-        <CustomLink title='Github' href='https://github.com/fedimint' />
-      </Flex>
+      <CustomLink
+        href={`https://github.com/fedimint/ui/releases/tag/${LATEST_RELEASE_TAG}`}
+      >
+        {LATEST_RELEASE_TAG}
+      </CustomLink>
+      <CustomLink href='https://fedimint.org'>&copy; Fedimint</CustomLink>
```
alexlwn123, line 47: nit: add locale

apps/router/src/context/AppContext.tsx
alexlwn123, line 1: 👏👏👏👏

alexlwn123 (approved):
Bravo 👏👏👏👏, Massive improvement to the look & feel.

Just for clarity, is it still possible to connect to multiple services from a single frontend? Ideally it should still be possible to connect to multiple services, even though most users will only add a single service.

If it's still possible, then let's ship it. If not, we should check with @elsirion if it's acceptable to ship as is.

alexlwn123:
Gotcha, makes perfect sense. Let's ship it!

apps/router/src/components/Footer.tsx
alexlwn123: actually nvm... my b

PR #623, opened 2025-02-19
Pass url as query string param

### LINKED ISSUE #579: Allow providing a guardian URL as URL parameter
Now that we have the central admin dashboard it would be nice to be able to deep-link into it such that a federation/gateway is automatically added to the list of services and opened: e.g. `https://admin.fedimint.org/#wss://myfederation.net` would add and open `myfederation.org`.

tsmith123:
Fixes #579 

This PR allows a user to pass a `url` query string param for a guardian/gateway when loading the UI. This string value will be added to the input box automatically.

**Note:** whilst we could connect automatically this would potentially cause a few issues such as if a user clicked back in the browser after connecting it would then connect them again. I prefer asking the user to click the connect button because this means they will follow the same flow as those who don't provide a query string param meaning we can reuse the same logic.

**Example**
```
https://fedimint-ui.com/?url=wss://btcpp-11.cypheru.net/ws/
```

Screenshot 2025-02-19 at 13 25 04

### DIFF
--- a/apps/router/src/hooks/custom/useQuery.tsx
+++ b/apps/router/src/hooks/custom/useQuery.tsx
@@ -1,0 +1,10 @@
+// from react-router-dom examples
+import { useMemo } from 'react';
+import { useLocation } from 'react-router-dom';
+
+export const useQuery = () => {
+  const { search } = useLocation();
+
+  // return () => new URLSearchParams(search);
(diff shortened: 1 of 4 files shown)


alexlwn123 (approved):
looks great

PR #626, opened 2025-02-27
Fix Vercel Refresh Issue

tsmith123:
When using the Vercel hosted UI and on any page other than root, if you hit refresh then you will get a 404. This PR attempts to solve the issue by updated the `vercel.json` file. I think this issue only applies to Vercel.


tsmith123:
Seems to bit an issue with the built version of the app as it happens locally too. Still investigating...

tsmith123:
So it appears to be a problem locally and on Vercel. Both seem to be fixed now in this PR.

alexlwn123 (approved):

Issue #628, opened 2025-03-01
The UI is confusing


dpc:
I'd like to fix:

https://github.com/fedimint/fedimint/issues/6934

I need a UI, so I go to https://github.com/fedimint/fedimint-ui/

I see https://ui-umber-ten.vercel.app/ , which looks like a hosted version. Great. I click.

It asks me for "Guardian URL". Hmmm... Which one is it exactly?

The host is `fedimintd-04.dev.fedimint.org`. I think it should be `wss://fedimintd-04.dev.fedimint.org`. I type that.

![Image](https://github.com/user-attachments/assets/9a1936f0-64e6-419f-aea4-366e3cf2c55d)

Hmmm... I'm quite sure the API is online... Am I wrong? Why doesn't it work?

Maybe I'm supposed to use `https://fedimintd-04.dev.fedimint.org`? I try that, I get a different error:

![Image](https://github.com/user-attachments/assets/2271e7f1-f565-4218-9284-a42d2d42fd5f)

Which, at least initially gave me impression that indeed `https://` might be the way to go.

Now I am guessing that when https is used, it assumes I want to interact with a gateway?

Anyway, the result is that I'm a Fedimint developer and I and can't accomplish my task, and I'm confused what's going. 

**Edit**: I think I heard that the UI is not compatible with 0.6 yet, so I'm guessing that's what is happening, but I think it's a good opportunity to think if we can eliminate some uncertainty here.

I think we should put more guardrails, be more specific about format of the URL we expect (the placeholder could suggest `wss://...` or maybe we should give an example somewhere next to the address box). 

We probably should also not try to be smart and have the user explicitly select if they want Fedimint Guardian UI or LN Gateway Admin UI, otherwise they might confuse themselves like I just did.

tsmith123:
Thanks for the feedback. Will restore the original placeholder to indicate that we're looking for a websocket url. The error seen is something that's existed for a while and it's on our radar to improve. I'm not sure about compatibility with 0.6 yet - something I need to dig into.

PR #630, opened 2025-03-01
Improve Placeholder On Homepage

tsmith123:
After some feedback in #628 this PR adds an example url as the placeholder on the Homepage to indicate it's a websocket url that's required.

Screenshot 2025-03-03 at 10 59 23


dpc (approved):

dpc (dismissed):
Wait. I think the port is wrong, and generally we shouldn't need to specify it anyway.

alexlwn123 (approved):

PR #633, opened 2025-03-08
Feat: Add toggle for privacy check for ValidateIcon 

### LINKED ISSUE #605: Fix validateIcon function for federation_icon_url
I noticed this behavior for the federation_icon_url meta field which is nice because it displays a little image box if the url is valid

However there is a discrepancy with the meta fields because technically `federation_icon_url` is not in the list of meta fields defined by Fedimint the protocol: https://github.com/fedimint/fedimint/tree/master/docs/meta_fields

So for the Fedi app we wanted to use a lot more meta fields and followed the [guidance](https://github.com/fedimint/fedimint/tree/master/docs/meta_fields#third-party-extensions) to namespace them with a `fedi:` prefix to future-proof them against collisions by other third party extensions

but `fedi:federation_icon_url` does not trigger the image preview so when making recommendations on how to manage meta for the best in-app experience, this creates some slight confusion/discrepancy

Image

Image

So I have 2 ideas

1) should federation_icon_url be officially defined by the Fedimint protocol @elsirion ? 
2) regardless of the above I think it'd be a nice improvement to instead of having the validation be triggered by a specific field name, it can simply detect if **any** Value field has an `https::` prefix + a `png/jpg/svg` suffix and trigger the image preview box. this would be useful for any other potential use of images in these meta fields so maybe that approach is cleaner?

MrImmortal09:
Closes #605 

Adds privacy feature to control when image URLs are validated and previewed in the custom metadata fields and sites input components. It adds a toggle switch that allows users to disable automatic external requests for image validation when privacy is a concern.

[Screencast from 2025-03-09 01-54-58.webm](https://github.com/user-attachments/assets/3283cd5a-dcca-46b7-92af-7a56366f3a27)

### DIFF
--- a/apps/router/src/guardian-ui/components/dashboard/tabs/meta/manager/SitesInput.tsx
+++ b/apps/router/src/guardian-ui/components/dashboard/tabs/meta/manager/SitesInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import {
   Flex,
   FormLabel,
@@ -6,8 +6,11 @@ import {
   Button,
   IconButton,
   Text,
+  Box,
+  Tooltip,
+  Switch,
 } from '@chakra-ui/react';
-import { FiX } from 'react-icons/fi';
+import { FiX, FiAlertTriangle, FiInfo } from 'react-icons/fi';
 import { IconPreview } from './IconPreview';
 
 interface Site {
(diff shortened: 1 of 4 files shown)


MrImmortal09:
The code now also displays an error message if there's an issue while fetching the image.

[Screencast from 2025-03-09 02-25-17.webm](https://github.com/user-attachments/assets/f4055880-59c5-4205-aa89-5a41a73c1d0f)

MrImmortal09:
@alexlwn123  , could you please review this?

apps/router/src/guardian-ui/components/dashboard/tabs/meta/manager/SitesInput.tsx
```diff
@@ -24,15 +27,61 @@ interface SitesInputProps {
...
   useEffect(() => {
     try {
       const parsedSites = JSON.parse(value);
       setSites(Array.isArray(parsedSites) ? parsedSites : []);
+
+      if (autoValidateUrls && Array.isArray(parsedSites)) {
+        parsedSites.forEach((site, index) => {
+          if (site.imageUrl) {
+            validateImage(site.imageUrl, index);
+          }
```
alexlwn123, line 78: Could we debounce this function to avoid making rapid calls if the user is typing? (e.g. only make try to validate 0.3 secs after the last keystroke)

image

alexlwn123 (changes requested):
Thanks for the contribution! Great PR overall.

Regarding the privacy toggle, I think [this comment on the issue](https://github.com/fedimint/fedimint-ui/issues/605#issuecomment-2626188937) is referring to the **user** fetching the federation icon (not the guardian). However, the toggle in the guardian-ui still seems like a nice-to-have, so I'm happy to leave it in.

Just address the one comment in the PR then it's good to merge.

MrImmortal09:
I've added debouncing to the SiteInput and Meta fields. Previously, two network calls were made—one for validation and another for fetching the image. Now, validation runs first, and if it fails, no additional call is made.

MrImmortal09:
@alexlwn123 I have made the changes, please review.

apps/router/package.json
```diff
@@ -28,6 +28,7 @@
     "react-dom": "^18.2.0",
     "react-icons": "^5.3.0",
     "react-minimal-pie-chart": "^8.4.0",
+    "use-debounce": "^6.0.0",
```
alexlwn123, line 31: let's not pull in an external dependency... this is simple enough to implement ourselves

alexlwn123 (changes requested):
Let's avoid the additional dependency. Here's a guide to make the debounce feature with timeouts:

https://medium.com/@sankalpa115/usedebounce-hook-in-react-2c71f02ff8d8

apps/router/package.json
MrImmortal09: I will implement that

MrImmortal09:
@alexlwn123 I have implemented the requested changes, please review

alexlwn123 (approved):
Thanks for the updates! Great contribution :)

PR #635, opened 2025-03-13
Handle New Bitcoin Status Response

tsmith123:
Fedimint version 0.6 returns a different response shape from the `check_bitcoin_status` endpoint and this makes it impossible to start the guardian ceremony. This PR handles responses from versions 0.5 and 0.6.

### DIFF
--- a/apps/router/src/api/GuardianApi.ts
+++ b/apps/router/src/api/GuardianApi.ts
@@ -2,6 +2,7 @@ import { JsonRpcError, JsonRpcWebsocket } from 'jsonrpc-client-websocket';
 import {
   AuditSummary,
   BitcoinRpcConnectionStatus,
+  BitcoinRpcConnectionStatusProgress,
   ClientConfig,
   ConfigGenParams,
   ConsensusState,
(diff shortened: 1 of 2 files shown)


apps/router/src/api/GuardianApi.ts
```diff
@@ -113,8 +114,19 @@ export class GuardianApi {
...
+  // This handles both 0.5 and 0.6 shaped responses and returns just 0.5 response (string)
+  checkBitcoinStatus = async (): Promise<BitcoinRpcConnectionStatus> => {
+    const result:
+      | BitcoinRpcConnectionStatus
+      | BitcoinRpcConnectionStatusProgress = await this.call(
+      SharedRpc.checkBitcoinStatus
+    );
+
+    if (typeof result === 'object') {
+      return result?.sync_percentage > 0.9999 ? 'Synced' : 'NotSynced';
```
tsmith123, line 126: The 0.9999 figure is one that dpc and I agreed upon to be a suitable threshold for now.

alexlwn123 (approved):
Nice, lgtm

PR #636, opened 2025-03-16
fix: check for service in storage

### LINKED ISSUE #629: Handle Missing LocalStorage
If for some reason the app is accessed when `service` isn't set in localstorage (typically when moving from v0.4 of UI to v0.5) then the app errors as the storage key is different. We should redirect the user back to the Login page to start the flow again.

tsmith123:
Fixes #629 

This PR ensures that a `service` exists in local storage when the app loads. If it doesn't then the user is redirected to the home page so that `service` can be set before continuing.


alexlwn123 (approved):
nice, lgtm

Issue #637, opened 2025-03-16
Add Error Boundary


tsmith123:
The app is currently lacking an ErrorBoundary. Makes sense to add one at the top level to catch anything unexpected.

Murdock9803:
@tsmith123 @alexlwn123 This is indeed a significant enhancement to be done. I would love to work on this, please assign the issue to me 🙌

tom-fedi:
Hey @Murdock9803 - sure, give it a go. In terms of "design", just something simple when such as the logo, some text saying something has gone wrong and a link back to the home page in the centre of the screen.

tsmith123:
Hey @Murdock9803 - sure, give it a go. In terms of "design", just something simple when such as the logo, some text saying something has gone wrong and a link back to the home page in the centre of the screen.

Murdock9803:
I got it @tsmith123 🙌 
Thanks for providing extra context, I was going to ask for the exact thing 😄

PR #638, opened 2025-03-19
Feat: Added a top-level Error to the App

Murdock9803:
Closes: #637 

### Description
The pull request aims to add a top level Error Boundary to the app, so that any errors are shown gracefully on the screen. This will prevent a White screen in case of errors in rendering, and will present a message screen with `Something went wrong` message.

### Screenshots
Screenshot 2025-03-19 084626

### Checklist
- [x] Written clean code to catch errors and present an error screen.
- [x] Error will be shown at the console.
- [x] Tested the solution to work locally.
- [x] Tried to keep the code-style consistent with the other files.


alexlwn123:
Hi @Murdock9803! Welcome & thanks for the contribution!

>  I was confused about what webSDK URL (wss://) to enter in the Homepage

Just to clarify, this is the [websocket](https://en.wikipedia.org/wiki/WebSocket) endpoint for the guardian server. The Fedimint Web SDK is something else

Will provide review in the PR

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,92 @@
...
+            fontSize='xl'
+            fontWeight='bold'
+            marginBottom={4}
+            marginTop={8}
+            color='red.500'
+          >
+            Something went wrong
+          </Text>
+
+          <a href='/'>
```
alexlwn123, line 81: the `a` tag is fine for the top-level route

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <Button colorScheme='blue'>Return to Home Page</Button>
```
alexlwn123, line 82: can we add an additional button to reload the current page? This will only be possible after moving the ErrorBoundary within the Route.

alexlwn123 (changes requested):
overall looks good. Left some comments to address before merging. 

I suggest trying to closely follow the existing patterns in the codebase already. Typically, if we're already doing something in the codebase, we'll want to do it the same way everywhere

apps/router/src/App.tsx
```diff
@@ -6,33 +6,36 @@ import { Guardian } from './guardian-ui/Guardian';
...
+              <Wrapper>
+                <GatewayContextProvider>
+                  <Gateway />
+                </GatewayContextProvider>
+              </Wrapper>
+            }
+          />
+        </Routes>
+      </Router>
+    </ErrorBoundary>
```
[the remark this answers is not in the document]
Murdock9803: @alexlwn123 I have put the ErrorBoundary in the `Wrapper.tsx` component, and it makes sure the error boundary is present in all the routes 🙌 Please suggest any further adjustments, if any 🙌

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,92 @@
...
...0.5881 24.4158 38.939 26.0649 38.939 28.0992C38.939 30.1335 40.5881 31.7826 42.6224 31.7826ZM35.5001 41.6474C35.5001 43.6817 33.851 45.3308 31.8167 45.3308C29.7824 45.3308 28.1333 43.6817 28.1333 41.6474C28.1333 39.6131 29.7824 37.964 31.8167 37.964C33.851 37.964 35.5001 39.6131 35.5001 41.6474ZM14.4892 45.3308C16.5234 45.3308 18.1726 43.6817 18.1726 41.6474C18.1726 39.6131 16.5234 37.964 14.4892 37.964C12.4549 37.964 10.8057 39.6131 10.8057 41.6474C10.8057 43.6817 12.4549 45.3308 14.4892 45.3308ZM7.36683 28.0992C7.36683 30.1335 5.71771 31.7826 3.68342 31.7826C1.64912 31.7826 0 30.1335 0 28.0992C0 26.0649 1.64912 24.4158 3.68342 24.4158C5.71771 24.4158 7.36683 26.0649 7.36683 28.0992ZM7.54018 14.8885C9.57447 14.8885 11.2236 13.2393 11.2236 11.205C11.2236 9.17074 9.57447 7.52163 7.54018 7.52163C5.50588 7.52163 3.85676 9.17074 3.85676 11.205C3.85676 13.2393 5.50588 14.8885 7.54018 14.8885ZM29.6066 18.5099L23.1529 15.4022L16.6992 18.5099L15.1051 25.4929L19.5717 31.0923H26.7342L31.2007 25.4929L29.6066 18.5099ZM23.1498 10.9372L23.1529 10.9357H23.1467L23.1498 10.9372ZM13.2077 15.7241L23.1498 10.9372L33.0919 15.7241L35.5496 26.4865L28.6688 35.1162H17.6309L10.7501 26.4865L13.2077 15.7241Z'
```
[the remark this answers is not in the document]
Murdock9803: @alexlwn123 I speculated this could be a problem, but since the error boundary was out of `Routes`, I couldn't include the <Logo /> SVG Component. The problem is solved now. Thanks for this 🙌

apps/router/src/components/ErrorBoundary.tsx
Murdock9803: I have included the `<Link>` as now our Error Boundary is at all the routes, and everything works fine. Please suggest if we should change it back to `a` 😄 🙌

apps/router/src/components/ErrorBoundary.tsx
Murdock9803: I have added the Reload button 🙌 The width for both the buttons is kept the same, for uniformity in design 🙌 
thanks for the review 🙌

Murdock9803:
Hello @alexlwn123 
I have made changes according to the reviews, and I'd be happy to make further changes, if required, to make the PR perfect 🙌 

<br/>
 
> Just to clarify, this is the [websocket](https://en.wikipedia.org/wiki/WebSocket) endpoint for the guardian server. The Fedimint Web SDK is something else.

Thanks for this, I will definitely read more about this and try to learn more about Fedimint too 🙌 My plan for few upcoming days is to read more and more about Fedimint and Bitcoin 🙌 

<br/>

> I suggest trying to closely follow the existing patterns in the codebase already. Typically, if we're already doing something in the codebase, we'll want to do it the same way everywhere.

Actually this is what I also aim for, but got confused with the `Routes` thing in this one. I will certainly remember to keep the code style consistent from now onwards 🙌 
Thanks for these reviews, I really appreciate it ☺️ 🙌

tsmith123:
Could you please add some notes on how you've tested this and also please update the initial screenshot now that it has changed 🙏

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  componentDidCatch(error: Error, errorInfo: React.ErrorInfo) {
+    console.error('Caught Error: ', error, errorInfo);
+  }
+
+  Reload = () => {
+    this.setState({ hasError: false });
+    window.location.reload();
+  };
```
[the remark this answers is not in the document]
Murdock9803: Actually I did not assign `const` with the function name, as it is a member of the Class. So it gave error.
Also I have removed the function entirely, as you suggested. `hasError` setting to `false` was really not needed, I should have looked upon it.. I'll try to be more careful next time 🙌

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+          <Link to={'/'}>
+            <Button width={52} colorScheme='blue'>
+              Return to Home Page
+            </Button>
+          </Link>
+          <Button
+            width={52}
+            marginTop={2}
+            onClick={this.Reload}
```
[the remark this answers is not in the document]
Murdock9803: I used `windows.location.reload` inside the onClick directly, but we can't use `useNavigate` hook as it is a class component (`ErrorBoundary`). Should I research more about what can be done ?

Murdock9803:
@tsmith123 Thanks for the review 🚀 
I have updated the screenshot after the recent changes, the current state of the Error Boundary screen.
I have also tried to implement your suggestions to make the code better 🙌 

> Could you please add some notes on how you've tested this

I tested this by implementing the following two steps:
1. bypassing the homepage screen.
2. creating an intentional error in `guardian UI` page.

To bypass the homepage, I changed `App.tsx` from `<Route path='/' element={<HomePage />} />` to ` <Route path="/" element={<Navigate to="/guardians/1" />} />`

To create an intentional error in `Guardian.tsx` : ` throw new Error("Error in Guardian component");`

This is how I tested the functioning and UI of the Error Boundary screen 🙌 Once again, thanks 🙌

tsmith123:
Thanks @Murdock9803 - will pull this down and give it a test myself when I get a chance.

apps/router/src/components/ErrorBoundary.tsx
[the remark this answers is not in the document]
tsmith123: Yeah ignore that suggestion about the `const` - I forgot that this was a class component. Don't see many of them around these days. 😆

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <Logo />
+          <Text
+            fontSize='xl'
+            fontWeight='bold'
+            marginBottom={4}
+            marginTop={8}
+            color='red.500'
+          >
+            Something went wrong!
+          </Text>
```
tsmith123, line 48: This would need a translation. Don't worry about doing the translations but you just need to add the en version and hook it up to use this string.

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,71 @@
...
+              Return to Home Page
+            </Button>
+          </Link>
+          <Button
+            width={52}
+            marginTop={2}
+            onClick={() => window.location.reload()}
+            colorScheme='blue'
+          >
+            Reload page
```
tsmith123, line 61: Translation

apps/router/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,71 @@
...
+            marginBottom={4}
+            marginTop={8}
+            color='red.500'
+          >
+            Something went wrong!
+          </Text>
+
+          <Link to={'/'}>
+            <Button width={52} colorScheme='blue'>
+              Return to Home Page
```
tsmith123, line 52: Translation

apps/router/src/components/ErrorBoundary.tsx
Murdock9803: Thanks for pointing this out 😄  I have added the translations 🙌

apps/router/src/components/ErrorBoundary.tsx
Murdock9803: Done 🙌

apps/router/src/components/ErrorBoundary.tsx
Murdock9803: Done 🙌

Murdock9803:
@tsmith123 
I have added the translations. At first, I thought it was straightforward but as the `ErrorBoundary` is a class component, it became somewhat different. I got to learn a new thing and the texts work fine in the screen 🙌

PR #661, opened 2025-07-12
Guardian authentication QR code missing password

kleysc:
Issue #653 

### Problem
Guardian QR auth was failing because of a sessionStorage key mismatch - code was looking for `'guardian-ui-key'` but password was stored under the dynamic `serviceConfigId`.

### Solution
Discussed with @otech47 and went with prop-passing instead of fixing storage keys.

Flow: `useGuardianApi()` → `FederationAdmin` → `DangerZone` → `GuardianAuthenticationCode`

### Changes
- Pass password through props instead of sessionStorage lookup
- Remove storage dependency
- Update interfaces for type safety

### Testing
1. Set up guardian federation
2. Click "Authenticate as Guardian" 
3. Verify password appears in QR code

### Verification
Password successfully included in QR authentication params
image

Issue #653, opened 2025-05-07
Password not encoded into the Authenticate as Guardian QR Code


otech47:
When I click to Authenticate as a Guardian I would expect the QR code to contain everything I need to be able to make admin API requests

![Image](https://github.com/user-attachments/assets/11b841c3-cf42-40bf-bc5e-808d54ddff89)

according to the logic the password should get included in the params before stringifying

![Image](https://github.com/user-attachments/assets/7639deda-2e57-4d35-976d-2284588853af)

but it doesn't actually get included, probably because in the session storage, I do not see any key for `guardian-ui-key` so the password cannot be retrieved to include

I do however see the value of the password stored in session storage, but the key is some long string which matches the ID stored in localstorage under the `fedimint_ui_state_v2` key under `service.config.id`

![Image](https://github.com/user-attachments/assets/1feb5db7-5db7-4f62-816a-e8f9a49521bb)

kleysc:
Hey @otech47, I've been looking into this and I think I have a possible solution. 

Instead of trying to fix the sessionStorage key mismatch, what if we just pass the password down through component props?

The flow would be: `useGuardianApi()` → `FederationAdmin` → `DangerZone` → `GuardianAuthenticationCode`

This way we skip the whole sessionStorage lookup and just get the password directly from the Guardian API hook. Feels cleaner and less error-prone to me.

I started implementing this approach and it seems to work well. The data flow becomes really straightforward.

Does this sound like a reasonable direction?

kleysc:
@otech47 #661 ready for review

tsmith123:
Qr code is also missing `url` but includes `inviteCode` which is not required.

PR #674, opened 2025-12-18
chore: add password and url to guardian qr code

tsmith123:
ref #653 
Builds on #661

- Adds `password` and `url` to Guardian authentication qr code
- Removes `inviteCode` from DangerZone as not needed


alexlwn123 (approved):
lgtm

